PR #7, opened 2017-02-28
Added .argv() to nconf so that it favors command line arguments befor…

Blackbaud-MichaelSpires:
…e the config file

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -86,7 +86,9 @@ logger.info('input_file: %s', JSON.stringify(context));
 /* Prepare configuration by initializing nconf, then passing that as the provider to the config object */
 const nconf = require('nconf');
 
-nconf.file(program.config_file);
+/* Favor command line arguments first then fallback on config file */
+nconf.argv()
+     .file(program.config_file);
 
 const config = require('auth0-extension-tools').config();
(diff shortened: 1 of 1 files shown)


mostekcm:
So, I merged before testing.  The problem is that we now have competing argv processors.  So, I'm not sure this really adds any value...  We could change it so that nconf runs the whole show...

Blackbaud-MichaelSpires:
Ah I see. I had tested it in a small proof of concept script but didn't see it being tested in the test files (must have missed it. Apologies).

What I'm trying to achieve is being able to pass in the client secret without having to write it to a config file. We have a build process that allows us to encrypt and hide the secret so that it is only revealed to the program during runtime and does not show up in logs or config files.

mostekcm:
ok, smart, let me whip something up

mostekcm:
@Blackbaud-MichaelSpires this is fixed in release 1.0.8, you can now set an environment variable to override anything, or pass -x for the secret

Blackbaud-MichaelSpires:
Awesome this is great. Thank you for the quick turnaround!

mostekcm:
no problem!

Issue #3, opened 2017-01-18
AUTH0_KEYWORD_REPLACE_MAPPINGS not working in database-connections files and rules


xurei:
I would like to configure a domain in the replace mappings to inject it in JS files (rules and custom database scripts). However, it doesn't seem to work.
Here is my config :
```
{
    "AUTH0_DOMAIN": "mydomain.auth0.com",
    "AUTH0_CLIENT_ID": "xxxxxxxxxxxxxxxxxxxxxxxxx",
    "AUTH0_CLIENT_SECRET": "xxxxxxxxxxxxxxxxxxxxx",
    "AUTH0_KEYWORD_REPLACE_MAPPINGS": {
        "PG_DOMAIN": "https://99db6f1c.ngrok.io"
    }
}
```

login.js :
```
function login(email, password, callback) {
	var gatewayURL = @@PG_DOMAIN@@;
	/* ... */
```

a-rule.js :
```
function getPGToken(user, context, callback) {
	var gatewayURL = @@PG_DOMAIN@@;
	/* ... */
```

In the Auth0 dashboard, the values are not replaced, hence the auth flow is broken.

Did I miss something or is it not implemented ?

mostekcm:
I'll check into this.  I'm not sure I had a test case for the .js files.  They should be implemented.  If not I will fix it.

mostekcm:
Once I have this PR merged and released, I can push a new release to auth0-deploy-cli: https://github.com/auth0-extensions/auth0-source-control-extension-tools/pull/8

mostekcm:
@xurei this is now fixed in version 1.0.4 on npm

nickcoury:
@mostekcm I'm still having issues getting this to work.  I looked through all the code and I believe this repo's code never gets the `mappings` object set with `AUTH0_KEYWORD_REPLACE_MAPPINGS`.

The constructor in `context.js` takes `mappings` as the second parameter, but in `index.js`, `context` is only constructed with the first parameter.

Additionally, `auth0-source-control-extension-tools` pulls that info from the config and passes it into `init()`, but again `context.js` ignores that argument.  See [here](https://github.com/auth0-extensions/auth0-source-control-extension-tools/blob/4a5089a4235d66ce51f225dd0769e33d2b30bde9/src/deploy.js#L45).

Am I missing something? I'm trying to replace a keyword in an HTML page and for the life of me can't get it to work as the documentation shows.

PR #9, opened 2017-06-20
Use mappings provided to context.init()

nickcoury:
This addresses a hole in Issue #3.  The keyword mappings are passed into `context.init()` but not used.  They are pulled from the constructor of `context` instead, but aren't passed in the actual script (just the tests).

This fix is backward compatible in case anyone is using the current setup.  If backwards compatibility isn't needed, this could replace the current mappings code.

### DIFF
--- a/src/context.js
+++ b/src/context.js
@@ -377,8 +377,10 @@ export default class {
     this.mappings = mappings;
   }
 
-  init() {
+  init(progress) {
     var me = this;
+    /* If mappings weren't provided, fall back to the ones provided to init() */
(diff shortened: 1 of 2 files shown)


mostekcm:
Great catch @nickcoury!  Not sure how I missed that!  Thanks!

mostekcm:
I will publish out to NPM shortly.  I want to take care of a couple of other bugs while I have it open.

mostekcm:
@nickcoury published to NPM as version 1.0.9

nickcoury:
Thanks! :+1:

PR #56, opened 2018-11-27
Ignore owners attribute for clients

### LINKED ISSUE #55: Bad Request when importing clients
Hello, I am trying to export from one tenant and import into a newly created one, both within the same private instance, and getting the following error:

`Bad Request: Payload validation error: 'Additional properties not allowed: owners'`

I'm guessing the Management API doesn't allow for custom fields? Is there any way of getting around this limitation?

Thanks

stevezau:
## ✏️ Changes
Ignore owners attribute for clients. This is set when you add admins that have access only to an application instead of the whole tenant. As it's specific to the tenant it should be ignored on dump/read-only.

  
## 🔗 References  
Fixes #55
  
## 🚀 Deployment
    
✅ This can be deployed any time

### DIFF
--- a/src/readonly.js
+++ b/src/readonly.js
@@ -20,7 +20,8 @@ const readOnly = {
     'global',
     'tenant',
     'custom_login_page_preview',
-    'config_route'
+    'config_route',
+    'owners'
   ]
 };
 


fyockm:
published as `auth0-deploy-cli@2.1.1`

PR #80, opened 2019-01-24
Fix for using the wrong thing for proxy

cherring:
## ✏️ Changes
  
We are trying to use the proxy when we push config to Auth0. It turns out that variable being used to try and do that is wrong. 

```
args.proxy_ul
```
is undefined, it looks like it should be 
```
params.proxy_url
```
So I have this PR in an attempt to not have to try and use a forked version of the library.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,7 +34,6 @@ async function run() {
   // Monkey Patch the superagent for proxy use
   const proxy = params.proxy_url;
   if (proxy) {
-    log.info(`PROXY: ${proxy}`)
     const proxyAgent = new HttpProxyAgent(proxy);
     const proxyAgentSsl = new HttpsProxyAgent(proxy);
     const OrigRequest = superagent.Request;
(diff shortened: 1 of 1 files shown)


cherring:
Sorry I will figure out the tests.

src/index.js
```diff
@@ -32,14 +32,15 @@ async function run() {
   }
 
   // Monkey Patch the superagent for proxy use
-  const proxy = args.proxy_url;
+  const proxy = params.proxy_url;
   if (proxy) {
+    log.info(`PROXY: ${proxy}`)
```
fyockm: Linting failed on this line. Please remove it and we can merge.

```
/home/circleci/repo/src/index.js
  37:32  error  Missing semicolon  semi
```

cherring: Hi @fyockm, stuff got busy for me and I got dragged away from this. I've made the change 👍

Issue #72, opened 2018-12-11
Export without secrets


giomaik:
Hello, 

Is it possible to export a yaml file stripping out the client secrets? It doesn't look like it. If not, it would be a great idea as we are looking to use an export of the current state before a deploy as an artifact for rollback and we wouldn't want it to contain sensitive info.

Thanks

stevezau:
I do like this idea. It's almost changing the option --strip to

`--strip-ids` and `--strip-sensitive`

@fyockm thoughts?

fyockm:
Yes, seems reasonable to me.

`--strip-ids` and `--strip-sensitive` make sense.  Just need to decide what all properties would be considered "sensitive".

omar:
Would it be fair to say that if you're trying to use this tool to create an authoritative configuration as code for your Auth0 setup that can be applied and promoted to different tenants (Dev, QA, Prod, etc.), you shouldn't export Auth0 settings with IDs or secrets?

If that's the case (and it is for our use case), then I'd encourage we change the CLI to behave that way by default and make export IDs and secrets an opt-in feature.

stevezau:
@omar i agree, will flag with team

PR #91, opened 2019-02-21
allow users to exclude properties on export

zxan1285:
## ✏️ Changes
  New `IGNORED_PROPERTIES` property added to config. It should allow customers to exclude certain props (such as `client_secret`) during export. 

Example:
```
"IGNORED_PROPERTIES": {
  "clients": [
    "client_secret"
  ]
}
```
  
## 🔗 References
  Github: #72 
  Slack: https://auth0.slack.com/archives/C9170558S/p1550745653014100
  
## 🎯 Testing

### DIFF
--- a/examples/directory/config.json.example
+++ b/examples/directory/config.json.example
@@ -10,5 +10,10 @@
   "AUTH0_EXCLUDED_RULES": [
     "rule-1-name",
     "rule-2-name"
-  ]
+  ],
+  "IGNORED_PROPERTIES": {
+    "clients": [
+      "client_secret"
+    ]
+  }
 }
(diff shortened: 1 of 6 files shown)


stevezau:
This is flexible but i think i prefer the suggestion by Omar https://github.com/auth0/auth0-deploy-cli/issues/72#issuecomment-461935706

That is make ids and secrets as an opt-in feature (`--export-ids`, `--export-secrets`).

I think that's a far better and more "secure" approach? thoughts?

fyockm:
for security reasons, I like the opt in. but could we also include the flexibility this solution allows? i.e. could we somehow support both?

zxan1285:
Closed in favor of #93

Issue #81, opened 2019-01-28
Directory export fails if entity name includes `/`


zamd:
In my tenant, I have a client named `AD / ADFS CUSTOM DOMAIN` and a0deploy failed with error as it's not escaping names properly.

```2019-01-28T11:46:09.280Z - debug: Error: ENOENT: no such file or directory, open 'z-exp/clients/AD / ADFS CUSTOM DOMAIN.json'
    at Object.openSync (fs.js:439:3)
    at Object.writeFileSync (fs.js:1190:35)
    at clients.forEach.client (/usr/local/lib/node_modules/auth0-deploy-cli/lib/context/directory/handlers/clients.js:49:23)
    at Array.forEach (<anonymous>)
    at Object.dump (/usr/local/lib/node_modules/auth0-deploy-cli/lib/context/directory/handlers/clients.js:46:11)
    at Promise.all.Object.entries.map (/usr/local/lib/node_modules/auth0-deploy-cli/lib/context/directory/index.js:87:36)
    at Array.map (<anonymous>)
    at _default.dump (/usr/local/lib/node_modules/auth0-deploy-cli/lib/context/directory/index.js:85:58)
2019-01-28T11:46:09.280Z - error: Problem running command export
2019-01-28T11:46:09.280Z - error: Problem exporting clients```

stevezau:
Fixe in #92

PR #92, opened 2019-02-21
Sanitize file and folder names

stevezau:
## ✏️ Changes
Property handle folder and file names. 

Fixes https://github.com/auth0/auth0-deploy-cli/issues/81


stevezau:
> @stevezau this looks great! could we add some tests?

Done.

PR #93, opened 2019-02-22
KEY-684: Export without secrets

### LINKED ISSUE #72: Export without secrets

zxan1285:
## ✏️ Changes
Fixes #72 
- `strip` option is replaced by `export_ids`. It's false by default, which means that IDs will be removed from exported object.
- `INCLUDED_PROPS` and `EXCLUDED_PROPS` are added to the config. Those can be used for flexible tweak of what properties should be included into exported objects.

## 🔗 References
  Jira: https://auth0team.atlassian.net/browse/KEY-684
  Github: #72
  
## 🎯 Testing 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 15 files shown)


zxan1285:
> @zxan1285 if you agree, we can just use #94

Sure.

PR #94, opened 2019-02-22
KEY-684: Export without secrets

### LINKED ISSUE #72: Export without secrets

fyockm:
## ✏️ Changes
Fixes #72 
- `--strip` option is replaced by `AUTH0_EXPORT_IDENTIFIERS: true`. It's now false by default, which means that IDs will be removed from exported object.
- `INCLUDED_PROPS` and `EXCLUDED_PROPS` are added to the config. Those can be used for flexible tweak of what properties should be included into exported objects.

## 🔗 References
  Jira: https://auth0team.atlassian.net/browse/KEY-684
  Github: #72
  Supersedes: #93 
  
## 🎯 Testing 

### DIFF
--- a/src/args.js
+++ b/src/args.js
@@ -58,20 +58,20 @@ export default yargs
       describe: 'The JSON configuration file.',
       type: 'string'
     },
-    strip: {
-      alias: 's',
-      describe: 'Strip the identifier field for each object type.',
-      type: 'boolean',
-      default: false
-    },
     secret: {
       alias: 'x',
       describe: 'The client secret, this allows you to encrypt the secret in your build configuration instead of storing it in a config file',
       type: 'string'
+    },
+    export_ids: {
+      alias: 'e',
+      describe: 'Export identifier field for each object type.',
+      type: 'boolean',
+      default: false
     }
   })
-  .example('$0 export -c config.json --strip -f yaml -o path/to/export', 'Dump Auth0 config to folder in YAML format')
-  .example('$0 export -c config.json --strip -f directory -o path/to/export', 'Dump Auth0 config to folder in directory format')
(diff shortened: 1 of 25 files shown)


src/args.js
```diff
@@ -58,20 +58,20 @@ export default yargs
...
-      alias: 's',
-      describe: 'Strip the identifier field for each object type.',
-      type: 'boolean',
-      default: false
-    },
     secret: {
       alias: 'x',
       describe: 'The client secret, this allows you to encrypt the secret in your build configuration instead of storing it in a config file',
       type: 'string'
+    },
```
zxan1285: Probably should remove this. If we're going to use `AUTH0_EXPORT_IDENTIFIERS: true`

zxan1285: And we decided to keep `export_ids` param. If set, it'll override config value.

zxan1285:
> Looks good to me except..
> 
> * I think we need an area which explains all the config options/cmd line options? The examples can ref to it? What do you think?
> * what about identifiers? I think we need to strip them by default? Similar to this, another PR??
> * I think we might want to being the config.json and cmd line params together, to me the cmdline params should just set the config, they should be the same?

1. https://auth0.com/docs/extensions/deploy-cli this seems like a nice place for all the explanations. Yet i'm not sure i'm the best person to explain anything :)

2. We do strip ids by default here: https://github.com/auth0/auth0-deploy-cli/pull/94/files#diff-f4f8238e49dd5edfd8bf15d981ed2802R109

3. Added that in the last commit. `--export_ids` will override config value.

fyockm:
published as `auth0-deploy-cli@3.0.0`

PR #108, opened 2019-03-15
Strict formatting (props order) for yaml dump

### LINKED ISSUE #61: If would be nice if names were at the top, when exporting.
When I create an export of an existing environment, in yaml, in some cases the 'name' attribute is a few attributes down. I am certain this is purely based on how the data is returned. However it would be a lot easier to navigate the resulting file if name, template, identifier and then id (in that order) were the first few elements (if they exist).

### LINKED ISSUE #82: Exporting a tenant multiple time should result in the same tenant.yml
Currently, running `a0deploy export -f yaml` multiple times does not result in the same `tenant.yml`. In specific, the multiple `emailTemplates` in `tenant.yml` are not in the same order after each export. 

This makes it really difficult to compare changes programmatically after exporting settings. Maybe these need to be sorted by value (i.e. `emailTemplates.template`'s value) so that there aren't blocks of config moving around unnecessarily.

I would assume this could apply to other multi-value properties too.

zxan1285:
## ✏️ Changes
Sorting assets by properties to make sure that order of those properties in yaml will be the same on each `dump` run.
Fixes #61  
Fixes #82 

## 🔗 References
Github issues: #61 and #82

## 🎯 Testing

### DIFF
--- a/src/context/yaml/index.js
+++ b/src/context/yaml/index.js
@@ -4,7 +4,7 @@ import path from 'path';
 import { loadFile, keywordReplace, Auth0 } from 'auth0-source-control-extension-tools';
 
 import log from '../../logger';
-import { isFile, toConfigFn, stripIdentifiers } from '../../utils';
+import { isFile, toConfigFn, stripIdentifiers, formatResults } from '../../utils';
 import handlers from './handlers';
 import cleanAssets from '../../readonly';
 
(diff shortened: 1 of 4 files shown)


fyockm:
published as `auth0-deploy-cli@3.1.2`

PR #111, opened 2019-04-02
Do not export secrets

zxan1285:
## ✏️ Changes
Deploy CLI can export `rules configs`, but it cannot extract values, which is leading to saving secrets as `"key": "******"`. Database connection secrets (`options.configuration`) can be extracted, but only as encrypted values. Importing these secrets without editing will override current values with new  encrypted values and can potentially break rules or db-scripts.
Proposed solution is to avoid exporting these secrets.

## 🔗 References
Jira SD: https://auth0team.atlassian.net/projects/ESD/queues/custom/321/ESD-82

## 🎯 Testing
✅This change has unit test coverage


fyockm:
published as `auth0-deploy-cli@3.2.0`

PR #122, opened 2019-05-22
DXEX-98: fix default-app client props on export

zxan1285:
## ✏️ Changes
`allowed_clients`, `allowed_logout_urls`, `allowed_origins` and `callbacks` properties of `client` cannot be imported as `null`. But for some reason the `Default App` client has all those defined as `null`. We can check those props and replace them with `[]` during export.

## 🔗 References
Jira: https://auth0team.atlassian.net/browse/DXEX-98

## 🎯 Testing


fyockm:
published as `3.3.1`

Issue #133, opened 2019-06-21
AUTH0_KEYWORD_REPLACE_MAPPINGS not working for emailTemplates, database connections, email providers, and pages when using yaml; possibly more.


victoru:
I was first alerted to this issue due to errors when trying to push pages that exist under an environment named path. e.g. `./pages/##ENV##/email_verify.html`
```
pages:
  - name: login
    enabled: true
    html: "./pages/##ENV##/login.html"
  - name: password_reset
    enabled: true
    html: "./pages/##ENV##/password_reset.html"
```

```
2019-06-21T04:44:10.801Z - error: Problem running command import
2019-06-21T04:44:10.801Z - error: Problem deploying pages
```

after temporarily hardcoding those values, I ran into more errors:
```
2019-06-21T05:30:55.237Z - error: Problem creating connections {"name":"somewhere-com"}
APIError: {"statusCode":400,"error":"Bad Request","message":"\"##GOOGLE_APPS_CLIENT_ID##\" is not a valid google oauth 2 client id, verify it in your Google API Console","errorCode":"invalid_body"}
```

this error only occurs because the values were invalid for that connection, I soon discovered other configuration was pushed successfully with unprocessed keyword mappings.

shushen:
This should have been fixed in the latest releases.

PR #134, opened 2019-06-21
JIT parsing for AUTH0_KEYWORD_REPLACE_MAPPINGS

victoru:
## ✏️ Changes

> I'm trying to use this tool with yaml config and all configuration properties provided by environment variables instead of a config file. `nconf` seems to always load configs loaded from environment variables as strings. 

I'm wasn't quite sure of the best way to do this, but I have updated the import lib to parse `AUTH0_KEYWORD_REPLACE_MAPPINGS` when it contains valid JSON string right before it is passed to the format contexts.

## 🔗 References

> https://github.com/auth0/auth0-deploy-cli/issues/133

## 🎯 Testing

> 1. export tenant configuration as yaml.
> 2. modify tenant.yaml to use keyword mappings in email provider, pages, and applications sections.
> 3. import tenant config and confirm unparsed keywords have not made it into the tenant configuration on auth0.

### DIFF
--- a/src/commands/import.js
+++ b/src/commands/import.js
@@ -42,7 +42,15 @@ export default async function deploy(params) {
   nconf.overrides(overrides);
 
   // Setup context and load
-  const context = await setupContext(nconf.get());
+  const context = await setupContext(function parseMappings(obj) {
+    if (typeof obj.AUTH0_KEYWORD_REPLACE_MAPPINGS === 'string') {
+      try {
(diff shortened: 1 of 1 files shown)


victoru:
I never felt comfortable with this implementation, so I have instead switched to using a fabulous [template merging tool](https://github.com/geofffranks/spruce) before running `a0deploy`.

andreasgrimm:
@victoru why did you close the PR? Shouldn't this be fixed?? 🤔

PR #161, opened 2019-09-27
ESD-2006: Remove non-functioning --verbose flag

faroceann:
## ✏️ Changes

Remove `--verbose flag`, since it has no functionality. Update docs to show support for `--debug`

## 🔗 References

https://auth0team.atlassian.net/browse/ESD-2006

## 🎯 Testing

Manually test via `a0deploy --help`

### DIFF
--- a/src/args.js
+++ b/src/args.js
@@ -2,8 +2,8 @@ import yargs from 'yargs';
 
 export default yargs
   .usage('Auth0 Deploy CLI')
-  .option('verbose', {
-    alias: 'v',
+  .option('debug', {
+    alias: 'd',
     describe: 'Dump extra debug information.',
     type: 'string',
     boolean: true,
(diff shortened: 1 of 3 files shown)


src/args.js
```diff
@@ -2,8 +2,8 @@ import yargs from 'yargs';
 
 export default yargs
   .usage('Auth0 Deploy CLI')
-  .option('verbose', {
-    alias: 'v',
+  .option('debug', {
```
lzychowski: I was under impression that there is already a debug option somewhere, do we really want to convert this?

faroceann: we have the `yargs` framework in place and it has useful features, e.g., `--help`.  however, we do need to check whether `--debug` is true through the `yargs` api, so I have updated.

PR #179, opened 2019-11-04
Add dump and deploy commands

nikhilsh:
## ✏️ Changes

`Import` and `export` are ambiguous commands as someone might think that import might be importing auth0's settings to your local machine, while export might be exporting to auth0. Dump and deploy are better names for what the actions actually do.

## 🎯 Testing

It is the same functionality as import and export and doesn't need to be tested

### DIFF
--- a/src/args.js
+++ b/src/args.js
@@ -70,7 +70,7 @@ export default yargs
       default: false
     }
   })
-.command('deploy', 'Deploy Configuration', {
+  .command('deploy', 'Deploy Configuration', {
     input_file: {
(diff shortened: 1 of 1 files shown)


shawnmclean:
@zxan1285 is this the same as https://github.com/auth0/auth0-deploy-cli/pull/183?
If so, we could close this in favor of that change.

Thanks @nikhilsh!

nikhilsh:
@zxan1285 @shawnmclean Thanks for getting the function in!

PR #188, opened 2019-12-11
Updated README.md with additional roles required from Auth0 CLI clien…

borrell:
## ✏️ Change

Updated the scopes outlined in the README.md file when creating a the client application by hand, rather than with the official extension.

## 🔗 References

Scopes required by the extension outlined here -> https://github.com/auth0-extensions/auth0-deploy-cli-extension/blob/master/webtask.json

## 🎯 Testing

n/a


shawnmclean:
Thanks for the changes!

Issue #182, opened 2019-11-06
Problem exporting roles


willhackett:
### Description

> Unable to iterate over `{}` returned by `roles` in `context.assets`.

https://github.com/auth0/auth0-deploy-cli/blob/master/src/context/yaml/handlers/roles.js#L12

Suggestion:

`context.assets.roles instanceof Array ? context.assets.roles : []`

I've noticed a few things come back as objects so the OR expression probably isn't the best in this case.

### Reproduction

- Run `a0deploy export -c config.json -f yaml -o . -d true`

![image](https://user-images.githubusercontent.com/12025460/68259309-9a757880-008d-11ea-9228-fcaabbca1c91.png)

### Environment

> Please provide the following:

- **Version of this library used:** 3.6.3

willhackett:
Let me know if you want a PR.

aagavin:
Also running into the same issue

TranquilMarmot:
We've run into this as well!

PR #194, opened 2020-01-07
Fix export failing if roles is empty

TranquilMarmot:
## ✏️ Changes

The roles API will return an empty object if there are no roles set up in the tenant. The expected return value is an array.

With these changes, the empty object will be ignored. Without these changes, the code continues and an exception is thrown:

```
TypeError: roles.forEach is not a function
   at Object.dump (/my_project/node_modules/auth0-deploy-cli/lib/context/directory/handlers/roles.js:51:9)
```

## 🔗 References

This fixes https://github.com/auth0/auth0-deploy-cli/issues/182

## 🎯 Testing

This can be tested by doing a dump against a tenant without any roles.

(are there integration tests for this package?)

### DIFF
--- a/src/context/directory/handlers/roles.js
+++ b/src/context/directory/handlers/roles.js
@@ -24,7 +24,9 @@ function parse(context) {
 
 async function dump(context) {
   const { roles } = context.assets;
-  if (!roles) return; // Skip, nothing to dump
+
+  // API returns an empty object if no grants are present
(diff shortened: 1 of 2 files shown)


TranquilMarmot:
We only fixed this for roles, but this may be the case in other spots as well. This code is reused in many places.

It seems like the API itself should return an empty array (or null) instead of an empty object in this case, but this at least fixes this library.

Issue #207, opened 2020-02-10
Connections YAML without enabled_clients is wiping out configured clients.


whynotavailable:
**Please do not report security vulnerabilities here**. The [Responsible Disclosure Program](https://auth0.com/whitehat) details the procedure for disclosing security issues.

**Thank you in advance for helping us to improve this library!** Please read through the template below and answer all relevant questions. Your additional work here is greatly appreciated and will help us respond as quickly as possible. For general support or usage questions, use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com/). Finally, to avoid duplicates, please search existing Issues before submitting one here.

By submitting an Issue to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

We like using the CLI for configuring our connections, however we manage the enabled_clients through an automation service. If we do not include enabled_clients in our connection it wipes out the configured clients. If we do include the node, only those included will remain wiping out the rest.

It's due to this line of code. https://github.com/auth0/auth0-deploy-cli/blob/master/src/context/yaml/handlers/connections.js#L50 Where it defaults to `[]` if the entry doesn't exist.

### Reproduction

1. Create a connection through the CLI without the enabled_clients node
2. Add a client through another means (like the admin console)
3. re-run the CLI
4. Your client no longer is configured.

### Environment

- Version used was auth0-deploy-cli@3.5.0
- linux/amd64 (it's a dockerized environment)

whynotavailable:
I made a mistake when doing my initial research. The issue in question is in the auth0-source-control-extension-tools repo. I have made a PR and linked this issue in that PR.

shushen:
@whynotavailable The deploy-cli is expecting the configuration JSON files to be the single source of truth. Therefore, when the `enabled_clients` is not present, `deploy-cli` has to assume the intention is to remove all `enabled_clients`.

I think the described behavior is indeed by design.

whynotavailable:
@shushen The only field that works this way is enabled_clients. Everything else is passed through as it with no changes or assumptions.

jshanm:
@shushen  
But is there anyway I can set all my clients should be enabled with this connection ?Without listing all the clients in the yaml.

bastianwegge:
[...]

PR #210, opened 2020-02-24
Updated YAML handler to fix issue with enabled clients

whynotavailable:
## ✏️ Changes

Fixed an issue where an empty enabled_clients node would wipe out configured clients.

I changed the projected enabled_clients to a separate object, and merged it with Object.assign. I needed to use Object.assign so that the key would not exist in the final object. This should not be a breaking change, as it only serves to keep what's in the management API, not change it.

## 🔗 References

This is in reference to https://github.com/auth0/auth0-deploy-cli/issues/207.

## 🎯 Testing

I have updated the unit tests to confirm the change in the code. I didn't see any integration tests, so I apologize for not running them. I also confirmed the linter was happy with the changes.

### DIFF
--- a/src/context/yaml/handlers/connections.js
+++ b/src/context/yaml/handlers/connections.js
@@ -41,19 +41,18 @@ async function dump(context) {
 
   const clients = context.assets.clients || [];
 
-  // nothing to do, set default if empty
   return {
     connections: connections.map((connection) => {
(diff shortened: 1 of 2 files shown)


whynotavailable:
I closed this as it wasn't the solution I was looking for. Sorry for the noise.

PR #220, opened 2020-03-31
FEAT: Configurable connections directory

suwyn:
## ✏️ Changes

Adds a new configuration variable `AUTH0_CONNECTIONS_DIRECTORY` that allows one to override the default directory (`/connections`) in which to source tenant connection configurations.

This allows you to have separate connection directories for each tenant which greatly improves the ability to be explicit on the tenant targets for each connection.   As it stands now, in order to target a single tenant for a connection, one must add the connection name to `AUTH0_EXCLUDED_CONNECTIONS` on every other tenant which is cumbersome and hard to maintain.  

Further we have found that connection configurations are very environment specific (keys, endpoints, etc) which leads to excessive bloat in tenant configuration definition files.   With this change, the connection files themselves can be bound to a specific environments.

## 🎯 Testing

### DIFF
--- a/src/context/directory/handlers/connections.js
+++ b/src/context/directory/handlers/connections.js
@@ -6,7 +6,8 @@ import log from '../../../logger';
 import { isFile, getFiles, existsMustBeDir, loadJSON, sanitize, ensureProp } from '../../../utils';
 
 function parse(context) {
-  const connectionsFolder = path.join(context.filePath, constants.CONNECTIONS_DIRECTORY);
+  const connectionDirectory = context.config.AUTH0_CONNECTIONS_DIRECTORY || constants.CONNECTIONS_DIRECTORY;
+  const connectionsFolder = path.join(context.filePath, connectionDirectory);
   if (!existsMustBeDir(connectionsFolder)) return { connections: [] }; // Skip
 
   const foundFiles = getFiles(connectionsFolder, [ '.json' ]);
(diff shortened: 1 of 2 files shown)


shawnmclean:
Hi @suwyn, could you share some insights into how you're working with multiple tenants in the same repository? What does that workflow look like?

suwyn:
Hi @shawnmclean - We manage 4 tenants with a single auth0 configuration repository, they are development, ci, staging, and production.    Each has a separate config.yml that define `AUTH0_KEYWORD_REPLACE_MAPPINGS` along with tenant specific Auth0 settings.   We use the directory format.   

We use development as the shared tenant for local development, ci to run automated tests against, staging for integration testing and user verification, and production for production systems.

The problem that this solves is our connections are very tenant specific and we have a very large number of them (inbound SAML integrations with our customers).   These Connections do not exist for development and ci but without this feature we're forced to deploy all connections to all environments so that creates a lot of cruft in Development and CI tenants and also in the configuration files as many of the connection settings are tenant specific (i.e. certs, endpoints).

Does that help you understand our use case?   

thanks for reviewing!

shawnmclean:
Thanks @suwyn! It does, are there any entities other than connections where they may not exist in all environments? I'm thinking if we need to implement something similar for other entities to keep consistency.

In either case, I'll merge this in and get it released. 

Thanks for your time in sharing your use case and the PR!

suwyn:
no problem @shawnmclean happy to contribute.   Connections are the only entity where we have seen large drift of the configuration across tenants so we don't have a use case where we would need to configure the directory sources for other entities.

I think consistency would be nice.   I tried to limit the amount of change in my PR so didn't make the others configurable, but it should be pretty simple.

thanks again

PR #232, opened 2020-05-02
[ESD-5880] Phone hook support

shawnmclean:
## ✏️ Changes

1. Validation for send phone message hook is supported via the `auth0-source-control-extension-tools` update.
2. Added Keep a change log lint and autogenerator package.
3. Cleaned up CHANGELOG.md.

## 🔗 References

https://auth0team.atlassian.net/browse/ESD-5880


package-lock.json
```diff
@@ -191,6 +191,47 @@
         }
       }
     },
+    "@brightcove/kacl": {
+      "version": "0.1.8",
+      "resolved": "https://a0us.jfrog.io/a0us/api/npm/npm/@brightcove/kacl/-/kacl-0.1.8.tgz",
```
shushen, line 196: Are we supposed to avoid private npm registers in open-source projects like this?

shawnmclean: No there shouldn't be, seems if you have an .npmrc in your home directory, `npm i` will attempt to use that registry, although none of our dependencies are coming from that registry.

Issue #231, opened 2020-04-30
Inconsistent import behavior between directory and yaml formats can lead to tenant being wiped of all configuration.


suwyn:
### Description

Importing a directory configuration where the directory is missing a directory for a handler results in the import deleting all resources for the handler on the tenant.

This behavior is different for a yaml configuration, where the absence of a handler key in the yaml will **not** delete resources for that handler.

This issue caused us problems when we deployed an empty directory to a tenant.  The result was all resources of that tenant were deleted.

Expected behavior would be for no resources to be deleted unless an empty directory for that resource was present in the directory being deployed.

### Reproduction

**Behavior with Directory format**
1. Export a configuration in directory format.
2. Delete the `resource-servers` folder (or any of the folders)
3. Import the directory back into the tenant.
4. You'll either see a message that the resources could be deleted, or they will be deleted, based on `AUTH0_ALLOW_DELETE`

**Behavior with YAML format**
1. Export a configuration in YAML format.
2. Delete the `resource-servers` key in `tenant.yaml`
3. Import the yaml back into the tenant.
4. The resources for the key that was removed are not processed for deletion.

The differences is observed in `context.load()` where the directory format returns an empty array for the asset `resourceServers: []` and the yaml format will return `resourceServers: undefined`

### Environment

- Version: `4.2.1`

ojkelly-vha:
I'd probably expect it would delete them (how else would you clean up after yourself)?

I think if this behaviour either way was behind a flag it would be helpful. But as it is, with a full GitOps workflow this would be the only way to remove anything from your tenant, so it needs to be possible.

shushen:
Fixed by #234. Thanks for the report and the PR!

PR #234, opened 2020-05-06
Don't default to empty resource collections if directory doesn't exist

suwyn:
## ✏️ Changes

The PR changes the defaults in the asset collection for a directory import.   It now matches the asset collection for yaml imported so that both formats are imported with consistent behavior.

Previously when importing a directory where a resource directory didn't exist, a0deploy would delete (or warn) any resources of that type that exist in the tenant.   This behavior is different from the yaml import where a0deploy will skip any resource types not defined in the yaml rather than deleting them.

## 🔗 References

Refer to https://github.com/auth0/auth0-deploy-cli/issues/231 for a longer description of the issue this PR fixes.

## 🎯 Testing

There are instructions to reproduce in the above referenced issue.

### DIFF
--- a/test/context/directory/context.test.js
+++ b/test/context/directory/context.test.js
@@ -15,11 +15,11 @@ describe('#directory context validation', () => {
     const context = new Context({ AUTH0_INPUT_FILE: dir });
     await context.load();
 
-    expect(context.assets.rules).to.deep.equal([]);
-    expect(context.assets.databases).to.deep.equal([]);
-    expect(context.assets.pages).to.deep.equal([]);
-    expect(context.assets.clients).to.deep.equal([]);
-    expect(context.assets.resourceServers).to.deep.equal([]);
+    Object.entries(context.assets).forEach(([ k, v ]) => {
+      if (typeof v === 'undefined') delete context.assets[k];
+    });
+
+    expect(context.assets).to.have.all.keys('exclude');
   });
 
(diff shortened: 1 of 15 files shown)


test/context/directory/context.test.js
```diff
@@ -15,11 +15,11 @@ describe('#directory context validation', () => {
...
 
-    expect(context.assets.rules).to.deep.equal([]);
-    expect(context.assets.databases).to.deep.equal([]);
-    expect(context.assets.pages).to.deep.equal([]);
-    expect(context.assets.clients).to.deep.equal([]);
-    expect(context.assets.resourceServers).to.deep.equal([]);
+    Object.entries(context.assets).forEach(([ k, v ]) => {
+      if (typeof v === 'undefined') delete context.assets[k];
+    });
+
```
suwyn, line 21: Rather than checking the value of each asset collection and delete any that are undefined.    The advantage of this over the previous tests is that if there is a new resource type added one would not have to update this test to ensure the correct behavior on that resource type since we iterate over all resource assets rather than hard coding the ones we want to check.

shushen:
@luisbritos Could you please review this PR? Since we are updating all dependencies and dropping support for Node<8, right now is a good opportunity to include breaking changes like this.

shushen:
@suwyn Thanks a lot for the contribution!

Issue #185, opened 2019-11-27
Unable to export / import emailProvider settings


chronotc:
### Description

Currently export command exports everything including credentials for most entities apart from emailProviders.

This is somewhat unexpected considering everything else can be kept in sync using `export` and `import` which is generally recommended when moving to codify auth0 settings.

This behaviour is only documented in `https://github.com/auth0/auth0-deploy-cli/blob/master/src/context/defaults.js` but would be helpful if this is surfaced in the README.md

### Environment

auth0-deploy-cli - v3.6.3

jbryson3:
We're also stumbling on this asymmetry. We have our credentials for our `ses` emailProvider checked into our private repo so that `a0deploy import` works as expected.

Typically the workflow for us goes like this:
1. `a0deply export`
2. Discard the changes made to email provider credentials
3. commit any changes as "snapshotting the tenant"
4. Merge feature branch to master
5. 'a0deploy import`

Ideally, getting rid of step #2 would allow for easier automation of the deploy process.

I see a couple of potential solutions:
- Environment/Config variable to opt out of exporting/importing the email providers
- When exporting configurations, do not change the values if set in current config for email provider credentials

jbryson3:
After a bit of debugging, I now understand the flaw in my proposed solutions: namely that during `export`, the current `tenant.yaml` file isn't read or used as an input. This means that under the current architecture there is no way to execute an `if` statement where the current `tenant.yaml` is used.

The problem is that there are fields that are essentially "write-only". They are usually sensitive in nature, and cannot be read through the API, but can be used to configure with POSTs or PUTs. I'm not sure how the `export` command should handle these "write-only" keys. I would think they need to be part of the configuration (config.json or tenant.yml) so that `a0deploy import` will work properly, and the values can be updated through the `a0deploy` tool. However, the `export` command has no current knowledge of those keys and can only set some dummy defaults (current behavior).

Paths forward may include:
1. Some intelligent merging of the current `tenant.yaml` and the values retrieved from the tenant calls
    - This seems like a brittle and use case specific solution. I like the simplicity of the current architecture
2. Move the sensitive emailProvider information into the `config.json` to be used in the `defaults.js` when exporting, and can be used as 'overrides' when importing
    - This pollutes the `config.json` a bit with specifics about the emailProviders, especially since the objects are polymorphic in nature (apikey vs accessKeyId/secretAccessKey vs password)

Any other ideas?

Issue #235, opened 2020-05-09
Default values for emailProvider cannot be excluded 


sobil:
### Description

The default values for credentials from `emailProviderDefaults` are added after EXCLUDED_PROPS are processed and cant be removed from the dump. This can result in incorrect values being included in a subsequent import if not removed or overwritten.

This behavior could lead to accidental configuration, but there is no way to disable the default behavior.

Refer to this code area to see where the defaults are being added: https://github.com/auth0/auth0-deploy-cli/blob/master/src/context/directory/handlers/emailProvider.js#L31
 
### Reproduction

This is normal and expected behavior as documented and can be reproduced with the usage exmples: https://github.com/auth0/auth0-deploy-cli#usage

### Environment

- **Version of this library used:v4.2.2
- **Version of the platform or framework used, if applicable: auth0 hosted pSaaS
- **Other relevant versions (language, server software, OS, browser): All
- **Other modules/plugins/libraries that might be involved: N/A

shushen:
Resolved by #236. Thanks for the great contribution @sobil!
We will be releasing a major release with this fix soon.

PR #236, opened 2020-05-09
Allow excluding default values for emailProvider with `AUTH0_EXCLUDED_DEFAULTS`

sobil:
## ✏️ Changes

Exporting default values can be valuable for discovering the format and schema of the format. It can however lead to unintentionally setting configuration with incorrect values in an already running environment.

This change adds a mechanism to prevent the default values being included in the configuration dump

It also changes the output for the default values to be useful with Environment Variables or AUTH0_KEYWORD_REPLACE_MAPPINGS.

## 🔗 References

Main Issue: https://github.com/auth0/auth0-deploy-cli/issues/235

Example of issue occurring for an Auth0 customer: https://support.auth0.com/tickets/00455714
Potentially part of a solution to resolve: https://github.com/auth0/auth0-deploy-cli/issues/185

## 🎯 Testing


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Allow exluding default values for emailProvider with `AUTH0_EXCLUDED_DEFAULTS`
```
trevorgk, line 10: typo: 
```suggestion
- Allow excluding default values for emailProvider with `AUTH0_EXCLUDED_DEFAULTS`
```

trevorgk: Also in the PR title

shushen:
> code should throw an error if the user is including explicitly values in INCLUDED_PROPS but these values are being excluded by AUTH0_EXCLUDED_DEFAULTS so the user knows that is an error with the configuration and the result wouldn't be as expected.

@luisbritos IMO this should not be required as there is no conflict.

- `AUTH0_EXCLUDED_DEFAULTS` specified the behaviour for the handlers to fill in defaults if the fields are not present from results returned from management API. This happens before the [`clean_asset()`](https://github.com/auth0/auth0-deploy-cli/blob/9d8a0b3d26cb276c51233d1bb00d9fb5a6933b00/src/context/yaml/index.js#L105) function is called to process `INCLUDED_PROPS` and `EXCLUDED_PROPS`. 
- `INCLUDED_PROPS` only applies when making an exception to include [properties that are usually stripped off](https://github.com/auth0/auth0-deploy-cli/blob/9d8a0b3d26cb276c51233d1bb00d9fb5a6933b00/src/readonly.js#L4)

Therefore, 
- if a field is specified in `EXCLUDED_PROPS`, it will be excluded no matter how `AUTH0_EXCLUDED_DEFAULTS` is configured, and
- if `emailProvider` is specified in `AUTH0_EXCLUDED_DEFAULTS`, or for any property that we might want to fill in with defaults in future, it's unlikely we will include them as part of the [properties to be stripped off](https://github.com/auth0/auth0-deploy-cli/blob/9d8a0b3d26cb276c51233d1bb00d9fb5a6933b00/src/readonly.js#L4), but in case we do, then they can still be controlled by `INCLUDED_PROPS`, independently of whether we fill them with defaults or not.

shushen:
@luisbritos I'm merging this PR - we can address any further concerns separately.

@sobil Many thanks for the awesome PR!

PR #245, opened 2020-05-26
tests: validate defaults handling in context tests

shushen:
## ✏️ Changes

Additional test cases to cover the support for `AUTH0_EXCLUDED_DEFAULTS` introduced in #236 

## 🔗 References

PR #236 

## 🎯 Testing

### DIFF
--- a/test/context/yaml/context.test.js
+++ b/test/context/yaml/context.test.js
@@ -96,7 +96,7 @@ describe('#YAML context validation', () => {
     expect(context.loadFile(script)).to.equal('// empty');
   });
 
-  it('should dump tenant.yaml', async () => {
+  it('should dump tenant.yaml with defaults', async () => {
     const dir = path.resolve(testDataDir, 'yaml', 'dump');
     cleanThenMkdir(dir);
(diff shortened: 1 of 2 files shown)


test/context/yaml/context.test.js
```diff
@@ -116,7 +116,80 @@ describe('#YAML context validation', () => {
...
+          custom_login_page_on: true,
+          name: 'Global Client'
+        }
+      ],
+      connections: [],
+      databases: [],
+      emailProvider: {
+        enabled: true,
+        name: 'smtp'
+      },
```
shawnmclean, line 192: So `AUTH0_EXCLUDED_DEFAULTS: [ 'emailProvider' ]` is excluding just `emailProvider.credentials` but not necessarily the whole object?

shushen: Since we are not getting `emailProvider.credentials` from management API, the behavior before `AUTH0_EXCLUDED_DEFAULTS` was to provide some defaults. This flag just disables the filling of these defaults.

PR #258, opened 2020-07-13
Validate AUTH0_DOMAIN in config

chrisscott:
## ✏️ Changes

 Lack of domain validation can result in less than helpful error messages that bubble up from node-auth0
 - Added a check for properly formed `AUTH0_DOMAIN` in `config.json`
 - Added test

## 🔗 References

n/a

## 🎯 Testing

Test added. Manually, test using an invalid `AUTH_DOMAIN` in `config.js`

### DIFF
--- a/src/context/index.js
+++ b/src/context/index.js
@@ -33,6 +33,10 @@ export default async function(config) {
     throw new Error(`The following parameters were missing. Please add them to your config.json or as an environment variable. ${JSON.stringify(errors)}`);
   }
 
+  if (config.AUTH0_DOMAIN.startsWith('http') || !config.AUTH0_DOMAIN.endsWith('auth0.com')) {
+    throw new Error("The AUTH0_DOMAIN in config.json needs to match your tenant's domain. This should be in the format '{TENANT NAME}.[us?|eu|au].auth0.com' and can be found in the Domain setting of the Application created for the deploy-cli.");
+  }
+
   let accessToken = config.AUTH0_ACCESS_TOKEN;
 
   if (!accessToken) {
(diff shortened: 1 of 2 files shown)


sobil:
@chrisscott , How would this affect customers on custom domains not using `auth0.com`?

If the error messaging is particularly painful perhaps another option could be to add a preflight check to make sure the domain is an Auth0 tenant and then return the same error... 🤔  I'm sure there are probably a few ways to do that, but the first that come to mind is checking the jwks endpoint.

chrisscott:
@sobil this wouldn't be affected by custom domains as the Management API Audience doesn't use the custom domain but the Auth0 domain.

sobil:
> @sobil this wouldn't be affected by custom domains as the Management API Audience doesn't use the custom domain but the Auth0 domain.

When using a private appliance that's doesn't appear to be the case.  At least the working config I am looking at has an `AUTH0_DOMAIN` that does not end in `auth0.com`

silviom:
> > @sobil this wouldn't be affected by custom domains as the Management API Audience doesn't use the custom domain but the Auth0 domain.
> 
> When using a private appliance that's doesn't appear to be the case. At least with the working config I am looking at here has an `AUTH0_DOMAIN` that does not end in `auth0.com`

That's true, not all PSaaS environments use an `.auth0.com` subdomain. For instance, some auth0 hosted customers and all customer hosted use their own domains.

chrisscott:
Thanks for pointing this out @sobil and than you for the confirmation @silviom! I'll update the PR to reflect this.

chrisscott:
Due to the above, I'm not sure there's much value in this PR now. I'm going to close this but happy to revive it and update it if there's interest.

PR #261, opened 2020-07-20
Fix deploy error_pages without html field

luisbritos:
## ✏️ Changes

Error pages haven't an HTML code file, but when assets are being created to process changes, loadFile(page.html) function is called, this generates an error.

This PRefixes that issue

## 🔗 References

https://auth0team.atlassian.net/browse/ESD-7637

## 🎯 Testing

### DIFF
--- a/src/context/yaml/handlers/pages.js
+++ b/src/context/yaml/handlers/pages.js
@@ -12,7 +12,7 @@ async function parse(context) {
     pages: [
       ...context.assets.pages.map(page => ({
         ...page,
-        html: context.loadFile(page.html)
+        html: page.html ? context.loadFile(page.html) : undefined
       }))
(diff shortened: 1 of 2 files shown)


src/context/yaml/handlers/pages.js
```diff
@@ -12,7 +12,7 @@ async function parse(context) {
     pages: [
       ...context.assets.pages.map(page => ({
         ...page,
-        html: context.loadFile(page.html)
+        html: page.html ? context.loadFile(page.html) : undefined
```
seejee, line 15: `null` would probably be a little more idiomatic here, or something like: 

```
{
 ...page
 ...!!page.html && { html: context.loadFile(page.html) }
}
```

PR #273, opened 2020-08-18
feat: terraform export/mirgration support

shushen:
## ✏️ Changes

This PR adds export and migration support for terraform.

As an Auth0 Customer, I should be able to:

✅ Export my Auth0 configurations from cloud to `terraform-provider-auth0`-compatible configurations, with at least equivalent support as auth0-deploy-cli exporting. 

✅ Convert my existing auth0-deploy-cli configuration files (`directory` or `yaml`) into  `terraform-provider-auth0`-compatible configurations

## 🔗 References

https://auth0.com/blog/use-terraform-to-manage-your-auth0-configuration/

## 🎯 Testing

- [ ] This change has unit test coverage

- [ ] This change has integration test coverage

- [ ] This change has been tested for performance

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -8,4 +8,8 @@ RUN apt-get update \
   && apt-get clean -y \
   && rm -rf /var/lib/apt/lists/*
 RUN echo "source /etc/profile.d/bash_completion.sh" >> ~/.bashrc
+RUN curl -fsL https://releases.hashicorp.com/terraform/0.12.29/terraform_0.12.29_linux_amd64.zip -o terraform_0.12.29_linux_amd64.zip \
+  && unzip terraform_0.12.29_linux_amd64.zip \
+  && mv terraform /usr/local/bin \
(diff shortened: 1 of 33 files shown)


stuarthadfieldglobality:
Hello. This would be fabulously useful. Has work here stopped or?

luisdavim:
Curious why was this closed. Will Terraform not be officially supported?

willvedd:
The [Auth0 Terraform Provider](https://github.com/auth0/terraform-provider-auth0) is now officially supported. While it supports management of Auth0 configuration, it is not meant to replace this tool; both have their pros/cons and are available to support your specific development workflows.

Regarding the functionality of this tool spitting out Terraform configuration, our team has been kicking around that idea but the idea still needs to be investigated further. This branch was closed before my time but IIRC, there were several technical challenges with it that made it a non-starter. 

We appreciate the feedback and if anyone else wants this functionality, please let us know! And more specifically, we'd like to know more details about your development workflows and the challenges you think this feature will solve.

Issue #277, opened 2020-09-14
examples and tests for guardian/policies.json are slightly wrong


smcl:
### Description

We wanted to use the auth0 deploy CLI to force MFA for all our applications, however we based our settings on the examples & tests, which are wrong. 

Firstly if I deploy our tenant settings using the same contents as the example `/guardian/policies.json` (which is `["all-applications"]`) I get the below result:

    $ a0deploy import -c config.json -i .
    2020-09-11T14:58:14.137Z - info: Processing directory .
    2020-09-11T14:58:14.157Z - info: Getting access token for (DELETED)/(OUR-TENANT).eu.auth0.com
    2020-09-11T14:58:14.221Z - error: Problem running command import
    2020-09-11T14:58:14.221Z - error: Schema validation failed loading [
        {
            "keyword": "type",
            "dataPath": ".guardianPolicies",
            "schemaPath": "#/properties/guardianPolicies/type",
            "params": {
                "type": "object"
            },
            "message": "should be object"
        }
    ]

It appears that the actual contents of policies.json should be the following:

    {
      "policies": [ 
        "all-applications"
      ]
    }

Additionally I believe some of the tests - in `/test/context/directory/guardianPolicies.test.js` - use the wrong value. They contain "all-application" (note: missing "s"). This should really be changed in case anyone refers to these in future and erroneously believes them to contain the right value. 

### Reproduction

Deploy auth0 settings for a tenant using `a0deploy import -i .` inside a directory with a `/guardian/policies.json` set to the following:

    [
      "all-applications"
    ]

### Environment

We're using auth0 deploy CLI v5.1.5 and node.js v14 under Alpine Linux.

shushen:
Fixed by #278

PR #278, opened 2020-09-14
Issue 277 - update example and tests for guardian/policies.json

smcl:
## ✏️ Changes

As described in the linked issue, a couple of files had some slightly incorrect values:

* `/examples/directory/guardian/policies.json` was incorrectly structured, so if you tried to use it you would get an error when running `a0deploy`
* `/examples/context/directory/guardianPolicies.test.js` incorrectly referred to the policy "all-application" instead of the correct "all-applications"

I changed both of these files and no others.

## 🔗 References

This fixes issue #277 

## 🎯 Testing

This doesn't change how the `a0deploy` tool itself works. It is a good idea to run the tests before and after this change to confirm nothing has changed.

### DIFF
--- a/examples/directory/guardian/policies.json
+++ b/examples/directory/guardian/policies.json
@@ -1,3 +1,5 @@
-[
-  "all-applications"
-]
+{
+  "policies": [
+    "all-applications"
+  ]
+}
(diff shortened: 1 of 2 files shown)


shushen (approved):

shushen:
Thanks for the PR!

PR #294, opened 2020-12-09
Support PERSIST_MAPPINGS to not replace the configs with defined variable substitutions.

kailashyogeshwar85:
## ✏️ Changes

When automating using auth0 cli we want to sync the changes to made to auth0 dashboard to local repo.If we have already defined the mappings in our client config or any resource it will be replaced when we export the tenant configs.

## Testing
`npm test`

## Code Changes
Allow to specify PERSIST_MAPPINGS: bool in env specific config file passed to a0deploy. 

## 🎯 Testing

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -4,6 +4,7 @@ import sanitizeName from 'sanitize-filename';
 import { loadFile } from 'auth0-source-control-extension-tools';
 import dotProp from 'dot-prop';
 import log from './logger';
+import context from './context';
 
 export function isDirectory(f) {
   try {
@@ -41,10 +42,19 @@ export function loadJSON(file, mappings) {
   }
 }
 
-export function dumpJSON(file, mappings) {
+export function dumpJSON(file, mappings, persistMappings) {
   try {
     log.info(`Writing ${file}`);
-    const jsonBody = JSON.stringify(mappings, null, 2);
+    let jsonBody = JSON.stringify(mappings, null, 2);
+
+    // merge both config if PERSIST_MAPPINGS=true
+    if (persistMappings && fs.existsSync(file)) {
+      const currentConfig = fs.readJSONSync(file);
+      const newConfig = JSON.parse(jsonBody);
+      jsonBody = { ...currentConfig, ...newConfig };
+      jsonBody = JSON.stringify(jsonBody, null, 2);
+    }
+
     fs.writeFileSync(
       file,
       jsonBody.endsWith('\n') ? jsonBody : `${jsonBody}\n`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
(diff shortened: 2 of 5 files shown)


src/utils.js
```diff
@@ -41,10 +42,19 @@ export function loadJSON(file, mappings) {
...
   try {
     log.info(`Writing ${file}`);
-    const jsonBody = JSON.stringify(mappings, null, 2);
+    let jsonBody = JSON.stringify(mappings, null, 2);
+
+    // merge both config if PERSIST_MAPPINGS=true
+    if (persistMappings && fs.existsSync(file)) {
+      const currentConfig = fs.readJSONSync(file);
+      const newConfig = JSON.parse(jsonBody);
+      jsonBody = { ...currentConfig, ...newConfig };
```
shushen, line 54: The merge appears to be naive and overriding at the top level of the JSON object.
If both the local configuration (`currentConfig`) and management API configuration (`newConfig`) are changing fields, we might lost the local configuration changes.

For example, 
```json
currentConfig = {
    "a1": {
        "b1": "current_b1_val",
        "b2": "newer_b2_val"
   }
}

newConfig = {
    "a1": {
        "b1": "newer_b1_val",
        "b2": "current_b2_val"
   }
}

jsonBody = {
    "a1": {
        "b1": "newer_b1_val",
        "b2": "current_b2_val"
   }
}
```

shushen (commented):
Hi @kailashyogeshwar85,

Many thanks for the PR.
I'm hesitate to add this bidirectional sync feature as proposed here. The deploy-cli as in its current design does not have the facility to properly support smart merge and will unable to reliably determine which of the local configuration or the one retrieved via Management API is newer.

The current design of the tool is assuming the `export` command is used to bootstrap the configurations, then the configurations will be used as the single source of truth to manage changes down the road. Therefore, syncing changes made from the dashboard is not a use case we are targeting for.

src/utils.js
zebkailash: @shushen local changes are by default overridden whenever we export the changes from upstream. In that case, also all the fields are overridden by the new params as well as the substitute expression as well which is not somewhat user will prefer to write every time he syncs the changes from upstream.

zebkailash: There is no way to keep the mappings user has written once and then wants to update for eg: **stag** tenant by exporting the new configuration. By direct merging, the user can keep its configuration in git and later on do a diff and merge if he thinks some fields are unwanted.

kailashyogeshwar85: Hi @shushen,

There is no way for the user to keep the mapping that he has written once for all the environments. Using git and then merging configurations will allow the user to select the fields which are merged using the above strategy rather than writing those expressions every time configurations are synced from upstream.

phillipuniverse:
> Therefore, syncing changes made from the dashboard is not a use case we are targeting for.

Why? Personally, the ability to export makes it easier for me to discover what additional options are available from the tenant UI and learn what configuration options map to what UI sections. I suppose this could also be handled with documentation but it's nice to see it as concretely as a git diff of changing local configs.

shushen:
Hi there,

I'm closing this PR as explained [earlier](https://github.com/auth0/auth0-deploy-cli/pull/294#pullrequestreview-549454730) for our reasoning. I understand the community may have a different opinion but unfortunately all things considered this is a feature we don't feel we can commit to support.

PR #298, opened 2020-12-11
chore: refactor client ID to name conversions

shushen:
## ✏️ Changes

Code refactor to reduce repeated code and there should be no functional changes.

## 🔗 References

n/a

## 🎯 Testing

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -185,10 +185,13 @@ export function clearClientArrays(client) {
   return client;
 }
 
-export function convertClientIdToName(clientId, knownClients) {
-  const clients = knownClients || [];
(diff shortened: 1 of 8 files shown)


src/utils.js
```diff
@@ -184,3 +184,15 @@ export function clearClientArrays(client) {
 
   return client;
 }
+
+export function convertClientIdToName(clientId, knownClients) {
+  const clients = knownClients || [];
```
shawnmclean, line 189: Just curious on why not use default params?

shawnmclean (approved):
LGTM!

src/utils.js
shushen: Great point. Improved in e5d9619.

PR #305, opened 2021-01-15
Remove all actions commits.

### LINKED ISSUE #302: v5.4.0 uses a git dependency
### Description

This package now pulls its `auth0` dependency from a branch in a git project (git://github.com/auth0/node-auth0.git#actionsManager), which makes it unsuitable for production use.

See: https://github.com/auth0/auth0-deploy-cli/blob/master/package.json#L30

This branch could change making different installations of this package behave differently (which is an undesirable trait for production use). I understand if this was an alpha version of the library, but does not feel like this should be done in release versions.

### Reproduction

I initially discovered this because our build environment pulls dependencies through a proxy, and it was blocking the installation of this package because we have not allowed the port that `git://` uses.

image

### Environment

- **Version of this library used:** 5.4.0

shawnmclean:
## ✏️ Changes

This PR aims to reset master back to pre-actions commits.

### Github

- Action related changes are moved to the `beta` branch and tracked under the PR #306.
- Removed `5.4.0` and `6.0.0` release.

### NPM

- Actions changes will be deployed under the `beta` tag on npm while this stream of changes will continue to live under `latest` tag on npm.
- Deprecate `6.0.0` and `5.4.0` on npm.

### Outcome

This should leave the last working production version to be 5.3.2 on both github and npm.

Resolves #302


seejee (approved):
Looks good!

Issue #303, opened 2021-01-07
Unable to deploy email template; 'verify_email_by_code'


Jelle-vz:
### Description

When adding the email template for 'verify_email_by_code' to the deployment directory, the pipleline fails, saying:

```bash
2021-01-06T13:04:19.197Z - debug: Error: Schema validation failed loading [
    {
        "keyword": "enum",
        "dataPath": ".emailTemplates[6].template",
        "schemaPath": "#/properties/emailTemplates/items/properties/template/enum",
        "params": {
            "allowedValues": [
                "verify_email",
                "reset_email",
                "welcome_email",
                "blocked_account",
                "stolen_credentials",
                "enrollment_email",
                "change_password",
                "password_reset",
                "mfa_oob_code"
            ]
        },
        "message": "should be equal to one of the allowed values"
    }
]
```

### Reproduction

* Add email template `verify_email_by_code.html` & `verify_email_by_code.json` to emails directory
* deploy to a tenant using the `auth0-deploy-cli` (using v6.0.0)
* Check debug log.

### Environment

- **"auth0-deploy-cli": "^6.0.0"**
- **Gitlab deployment pipeline**

pmalouin:
Thanks for reporting @Jelle-vz and sorry for the experience. I'm working on addressing this.

In the mean time, you can skip that template on import by manually removing it from the YAML file or the /emails directory.

Jelle-vz:
HI @pmalouin, 
Thats great! Sure, we copy/paste the template manually now so we are not (really) blocked haha.

Do you have any indication on when we can expect the cli-tool to be updated?

Cheers!

pmalouin:
> Do you have any indication on when we can expect the cli-tool to be updated?

Things are in motion, but we have some quality review cycles to go through before we publish it. It can go fast if I can get some eyes on this, but could take a few days if people are busy.

pmalouin:
@Jelle-vz Fix is available starting with `auth0-deploy-cli@5.5.0+` ✔️ thanks for your patience

Jelle-vz:
Thanks!

PR #309, opened 2021-01-19
[CAUTH-837] add `verify_email_by_code` email template

pmalouin:
## ✏️ Changes

⚠️ Depends on: https://github.com/auth0-extensions/auth0-source-control-extension-tools/pull/123

⚠️ I'd like this change to be merged to `master`, I don't need this on the `beta` branch. Let me know if there's something to adjust here.

Add a new email template key `verify_email_by_code`, to support the "Verification Email (using Code)" template that was introduced in the past few months in the dashboard:
image

Fixes https://github.com/auth0/auth0-deploy-cli/issues/303

## 🔗 References

https://auth0team.atlassian.net/browse/CAUTH-837

## 🎯 Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "auth0": "^2.27.0",
     "auth0-extension-tools": "^1.4.4",
-    "auth0-source-control-extension-tools": "^4.1.12",
+    "auth0-source-control-extension-tools": "~4.3.0",
     "dot-prop": "^5.2.0",
     "fs-extra": "^7.0.0",
     "global-agent": "^2.1.12",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth0-deploy-cli",
-  "version": "5.3.2",
+  "version": "5.4.0",
   "description": "A command line tool for deploying updates to your Auth0 tenant",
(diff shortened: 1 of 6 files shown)


joseluisdiaz (approved):

package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "auth0": "^2.27.0",
     "auth0-extension-tools": "^1.4.4",
-    "auth0-source-control-extension-tools": "^4.1.12",
+    "auth0-source-control-extension-tools": "~4.3.0",
```
pmalouin, line 32: Note to maintainers: back in July, I found that setting this matcher to a caret (`^`) can be very risky for existing customers, because adding new features to `auth0-source-control-extension-tools` (semver minor) can be breaking for `auth0-deploy-cli`. Here, I'm reverting back to tilde matching (`~`) to make it safer to deploy CLI consumers in the future.

I don't think there's a simple way to automate a check for this and annoyingly, `npm` defaults to caret. We'll need to rely on code reviews to detect this...

pmalouin: BTW, 4.2.0 was never published, that's why the minor version skips to `3`...

jpadilla (approved):

PR #363, opened 2021-07-07
Implement prompts API support for settings

### LINKED ISSUE #229: Prompts API support
Do you planing add supporting to enable new login experience? I see API already exists https://auth0.com/docs/api/management/v2/#!/Prompts/get_prompts but there no interface for using this API in deploy cli.

Thank you in advance for your response!

maastiff:
## ✏️ Changes

This pull request aims to add support for prompts settings.

The "tools" part for prompts settings was already implemented to load / save configuration through API V2 which explain the need in read:prompts scope as stated in documentation.
This pull request implement the "context" side of prompts settings management which aims to dump / load data from json directory or yaml source.

see: https://auth0.com/docs/api/management/v2/#!/Prompts/get_prompts

## 🔗 References

Fixes #229 

## 🎯 Testing

Unit tests has been implemented.
You can try it out on a real Auth0 tenant the following way:

- run **export cli** on existing freshly created tenant in **directory output mode**
- you should retrieve a `prompts.json` file at root of output directory which should contains the following configuration 

```json
{
  "universal_login_experience":"classic",
  "identifier_first":false
}
```

- Go to the "Universal Login" settings page on your tenant, and ensure you see **classic mode enabled** (which is coherent with previously retrieved configuration)
https://manage.auth0.com/dashboard/<region>/<tenant_name>/login_settings

- Update the `prompts.json` file with the following content which aims to enable new universal login experience with identifier first flow

```
{
  "universal_login_experience": "new",
  "identifier_first": true,
  "webauthn_platform_first_factor": false
}
```

- run **import cli** on the same tenant in **directory mode**

- Go to the "Universal Login" settings page on your tenant, and ensure you see **new mode enabled** (which is coherent with previously updated configuration)
https://manage.auth0.com/dashboard/<region>/<tenant_name>/login_settings
- Then go to the "Authentication Profile" settings page on your tenant, and ensure you see "identifier first" enabled

Repeat the same operation in **yaml output mode**

### DIFF
--- a/src/context/directory/handlers/prompts.js
+++ b/src/context/directory/handlers/prompts.js
@@ -1,0 +1,37 @@
+import path from 'path';
+import {
+  existsMustBeDir, isFile, dumpJSON, loadJSON
+} from '../../../utils';
+
+function parse(context) {
+  const baseFolder = path.join(context.filePath);
+  if (!existsMustBeDir(baseFolder)) return {}; // Skip
+
+  const promptsFile = path.join(baseFolder, 'prompts.json');
+
+  if (isFile(promptsFile)) {
+    /* eslint-disable camelcase */
+    const prompts = loadJSON(promptsFile, context.mappings);
+
+    return {
+      prompts: prompts
+    };
+    /* eslint-enable camelcase */
+  }
+
+  return {};
+}
+
+async function dump(context) {
+  const { prompts } = context.assets;
+
+  if (!prompts) return; // Skip, nothing to dump
+
+  const promptsFile = path.join(context.filePath, 'prompts.json');
+  dumpJSON(promptsFile, prompts);
+}
+
+export default {
+  parse,
+  dump
+};
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Supported Auth0 Management API resources
 - [ ] [Log Streams](https://auth0.com/docs/api/management/v2#!/Log_Streams/get_log_streams)
 - [ ] [Logs](https://auth0.com/docs/api/management/v2#!/Logs/get_logs)
 - [x] [Organizations](https://auth0.com/docs/api/management/v2#!/Organizations/get_organizations)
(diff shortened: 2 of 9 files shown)


atambo:
This would be super useful, @shushen do you think this could be merged?

kahuna-celsius (changes requested):
This only addresses the Get/Patch on prompts settings resources. 
Not addressing Get/Put custom prompt text at all.

So it's not ok to tick the `Prompts` box as supported, unless everything is supported.

Idea:

During dump:
- Get all supported locales from tenant settings 
- Create a directory for every locale
- Create text JSON files for every locale/prompt combination and structure them into those directories

During deploy:
Iterate over the locale directories then over prompts in those directories and update every JSON file containing prompt texts present using
Call `PUT /api/v2/prompts/{promptName}/custom-text/{language_code}`

src/context/directory/handlers/prompts.js
```diff
@@ -0,0 +1,37 @@
...
+
+  return {};
+}
+
+async function dump(context) {
+  const { prompts } = context.assets;
+
+  if (!prompts) return; // Skip, nothing to dump
+
+  const promptsFile = path.join(context.filePath, 'prompts.json');
```
shushen, line 30: The `dump` operation doesn't work for me as it doesn't seem to create the directory for prompts.
Please see `https://github.com/auth0/auth0-deploy-cli/blob/d33413a789864af1ae345ef5debac0e24111d3ff/src/context/directory/handlers/clients.js#L42` for an example.

shushen (changes requested):
Many thanks for the PR but the export doesn't seem to be working, please see my comment.

willvedd:
Thanks for opening this PR! Looks like it's become stale so I'm opting to close. However, please note that providing feature parity is an important initiative that the team is planning for in the weeks ahead. I consider prompts to fall into that initiative. Also please note that we have #229 as an open issue so it's still being tracked somewhere.

PR #364, opened 2021-07-12
[DXEX-1527] Add action.runtime

turcottedanny:
## ✏️ Changes

Expose action.runtime to match APIv2. 

## 🔗 References

https://auth0team.atlassian.net/browse/DXEX-1527

## 🎯 Testing

Using export and import and validate that the action.runtime property is correctly set.

### DIFF
--- a/src/tools/auth0/handlers/actions.js
+++ b/src/tools/auth0/handlers/actions.js
@@ -14,6 +14,7 @@ export const schema = {
     additionalProperties: false,
     properties: {
       code: { type: 'string', default: '' },
+      runtime: { type: 'string' },
       dependencies: {
         type: 'array',
         items: {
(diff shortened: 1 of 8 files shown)


luisbritos (approved):
LGTM

src/tools/auth0/handlers/actions.js
```diff
@@ -153,6 +154,10 @@ export default class ActionHandler extends DefaultHandler {
         actionChanges.code = action.code;
       }
 
+      if (action.runtime !== found.runtime) {
```
shushen, line 157: This is an existing bug: the `actionChanges()` function doesn't seem to be called at all and the actions handler always update and redeploy all existing actions (thus creating a new version every time deploy-cli runs import).

This also leads to where if one specifies the `runtime` property and then remove it from the configuration, it does not revert to the default runtime (because the update operation is using the PATCH endpoint).

I don't think this needs be fixed in this PR but would like to surface the issue. I'll file a Jira ticket.

shushen (approved):

Issue #367, opened 2021-07-23
customScripts value not respected


skukx:
### Description

I'm currently using a0deploy (7.1.1) to apply changes to a development tenant. When altering paths to the
`customScripts` object, I've notice that the path the the script file is not respected.

```json
{
  "options": {
     ...
     "customScripts": {
        "change_password": "./change_password.js",
        "get_user": "./get_user.js",
        "login": "./another/path/login.js"
    },
    ...
  }
}
```

When applying the changes, the result is successful. However when checking the tenant through the auth0 dashboard, the original (`./login.js`) script still shows and not the `./another/path/login.js` script file.

The context being used is `Directory`. Again, the deployment succeeds overall, however the different path for login script is not used.

Thank you for your time and please let me know if I'm doing something wrong.

### Reproduction

**config.json**
```json
{
  "AUTH0_DOMAIN": "<insert>.us.auth0.com",
  "AUTH0_CLIENT_ID": "<client_id>",
  "AUTH0_CLIENT_SECRET": "<client_secret>",
  "AUTH0_ALLOW_DELETE": true,
  "AUTH0_KEYWORD_REPLACE_MAPPINGS": {
    "AUTH0_DOMAIN": "<tenant>",
    "AUTH0_CLIENT_ID": "<client_id>",
    "DB_LOGIN_SCRIPT_PATH": "./another/path"
  }
}
```

**database.json**
```json
{
  "options": {
    "mfa": {
      "active": true,
      "return_enroll_settings": true
    },
    "import_mode": false,
    "customScripts": {
      "change_password": "./change_password.js",
      "get_user": "./get_user.js",
      "login": "##DB_LOGIN_SCRIPT_PATH##/login.js"
    },
    "disable_signup": false,
    "passwordPolicy": "good",
    "password_history": {
      "size": 5,
      "enable": false
    },
    "strategy_version": 2,
    "requires_username": false,
    "password_dictionary": {
      "enable": false,
      "dictionary": []
    },
    "brute_force_protection": true,
    "password_no_personal_info": {
      "enable": false
    },
    "password_complexity_options": {
      "min_length": 12
    },
    "enabledDatabaseCustomization": true,
    "configuration": {}
  },
  "strategy": "auth0",
  "name": "My-DB",
  "is_domain_connection": false,
  "realms": [
    "My-DB"
  ],
  "enabled_clients": []
}
```

### Environment

> Please provide the following:

- **Version of this library used:** `7.1.1`

PR #368, opened 2021-07-23
Issue#367 - Load file configured in customScripts

skukx:
The customScripts in the directory context is not loading the configured
script paths. Rather it is loading the default paths within the
directory. To maintain configurability, update the getDatbase function
to load the path from the configured path location defined in
customScripts.

## ✏️ Changes

This applies a similar pattern done with the [YAML context](https://github.com/auth0/auth0-deploy-cli/blob/614b5158011b33686aaf28d8e963b2bc5f6cef74/src/context/yaml/handlers/databases.js#L19-L22)

- Update the `getDatabase` function within the databases module
- Change how custom scripts are loaded. Rather than loading files within the root of the folder, respect the path configured within the `customScripts` object.
- Rather than rejecting the file by name, only reject invalid properties within `customScripts` object.
- Updated tests because no matter the configured script path in the main config, the same file was loaded every time. This meant tests would fail after the change due to the file not exiting when trying to load.

## 🔗 References

Issue #367 

## 🎯 Testing

This change is covered by unit tests already. I've updated some names in the test to demonstrate custom named scripts. If the configured path or file doesn't exist then the script should fail.

### DIFF
--- a/src/context/directory/handlers/databases.js
+++ b/src/context/directory/handlers/databases.js
@@ -8,7 +8,6 @@ import {
   existsMustBeDir,
   dumpJSON,
   loadJSON,
-  getFiles,
   sanitize,
   mapClientID2NameSorted
 } from '../../../utils';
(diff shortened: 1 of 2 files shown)


skukx:
I've noticed the database object being created is referencing a property that no longer exists `metaData.customScripts` vs `metaData.options.customScripts`. I'm assuming this is for backward compatibility?
https://github.com/auth0/auth0-deploy-cli/blob/614b5158011b33686aaf28d8e963b2bc5f6cef74/src/context/directory/handlers/databases.js#L38-L39

atambo:
What I'm seeing is that only the first database script (login) is being imported into auth0. All of the other scripts are not being updated. @ggoodman you look like the last auth0 person who has touched the repo. Could you maybe take a look at this?

shushen (approved):
Thanks for the PR!

neoeinstein:
Note that this is a _breaking change_. It just caused us to deploy to production in a way that actually _deleted_ all of our database connections, resulting in a production outage until we reverted back to a previous version. Prior behavior did _not_ require a `customScripts` element, but used sibling files with the appropriate name. This change _removed_ that functionality.

PR #369, opened 2021-08-12
[IDS-3074] Updated structure when dumping orgs

mcastany:
## ✏️ Changes

Fixes an issue when exporting organizations as a directory, connections are not structured in the right way, causing the import to remove any connection on the organizations

## 🎯 Testing

🚫  This change has integration test coverage

### DIFF
--- a/src/context/directory/handlers/organizations.js
+++ b/src/context/directory/handlers/organizations.js
@@ -35,6 +35,21 @@ async function dump(context) {
   organizations.forEach((organization) => {
     const organizationFile = path.join(organizationsFolder, sanitize(`${organization.name}.json`));
     log.info(`Writing ${organizationFile}`);
+
+    if (organization.connections.length > 0) {
+      organization.connections = organization.connections.map((c) => {
+        // connection is a computed field
+        const name = c.connection && c.connection.name;
+        delete c.connection_id;
+        delete c.connection;
+
+        return {
+          name,
+          ...c
+        };
+      });
+    }
+
     dumpJSON(organizationFile, organization);
   });
 }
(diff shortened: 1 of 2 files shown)


mcastany:
Demo testing the flow with the change included 

https://user-images.githubusercontent.com/1765468/129265310-1dcbd03a-6fef-41c8-8ac0-b44e5a61bd3a.mov

src/context/directory/handlers/organizations.js
```diff
@@ -35,6 +35,21 @@ async function dump(context) {
   organizations.forEach((organization) => {
     const organizationFile = path.join(organizationsFolder, sanitize(`${organization.name}.json`));
     log.info(`Writing ${organizationFile}`);
+
+    if (organization.connections.length > 0) {
+      organization.connections = organization.connections.map((c) => {
+        // connection is a computed field
```
jhiner, line 41: What does "computed field" mean?

mcastany: "Computed fields are virtual values or objects that are dynamically computed"

In this case, the connection object inside enabled connections is not stored in the enabled_connections object, is a copy of what the connection has. That's the reason why we are removing them. The import commands gets the complete object and tries to recreate it in Auth0. Sending connection would fail

dctoon (approved):

src/context/directory/handlers/organizations.js
```diff
@@ -35,6 +35,21 @@ async function dump(context) {
...
+        const name = c.connection && c.connection.name;
+        delete c.connection_id;
+        delete c.connection;
+
+        return {
+          name,
+          ...c
+        };
```
ggoodman, line 49: What do you think about this as a means to avoid mutating the referenced object in the mapper?

```suggestion
        const name = c.connection && c.connection.name;

        return {
          name,
          ...c
          connection_id: undefined,
          connection: undefined,
        };
```

dctoon (approved):

ggoodman (approved):

PR #370, opened 2021-08-19
[DXEX-1721] Fix client metadata property deletion

ggoodman:
## Description

Fixes an inconsistency between how we calculate changes on deep metadata-like objects and with how APIv2 expects such changes to be expressed when a property is deleted.

### DIFF
--- a/src/tools/utils.js
+++ b/src/tools/utils.js
@@ -62,7 +62,7 @@ export function dumpJSON(obj, spacing = 0) {
   return JSON.stringify(obj, null, spacing);
 }
 
-export function calcChanges(assets, existing, identifiers = [ 'id', 'name' ]) {
+export function calcChanges(assets, existing, identifiers = [ 'id', 'name' ], objectFields = []) {
   // Calculate the changes required between two sets of assets.
   const update = [];
   let del = [ ...existing ];
@@ -108,7 +108,13 @@ export function calcChanges(assets, existing, identifiers = [ 'id', 'name' ]) {
               if (found[i]) obj[i] = found[i];
               return obj;
             }, {}),
-            ...asset
+            // If we have any object fields, we need to make sure that they get
+            // special treatment. When different metadata objects are passed to APIv2
+            // properties must explicitly be marked for deletion by indicating a `null`
+            // value.
+            ...(objectFields.length
+              ? processChangedObjectFields(asset, found, objectFields)
+              : asset),
           });
         }
       }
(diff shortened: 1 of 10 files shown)


ggoodman:
Note to reviewer: I have another PR to follow this to fix some yaml serialization. Once that also lands I'll do the changelog and release as a separate PR.

src/tools/utils.js
```diff
@@ -62,7 +62,71 @@ export function dumpJSON(obj, spacing = 0) {
...
+/**
+ * @template T
+ * @param {typeof import('./auth0/handlers/default').default} assetType
+ * @param {T} desiredAssetState
+ * @param {T} currentAssetState
+ * @param {string[]} [objectFields=[]]
+ * @param {boolean} [allowDelete=false]
+ * @returns T
+ */
+function processChangedObjectFields(assetType, desiredAssetState, currentAssetState, objectFields = [], allowDelete = false) {
```
turcottedanny, line 74: Can we rename assetType to handler, for me assetType is inferring a string.

test/tools/utils.tests.js
```diff
@@ -159,20 +178,146 @@ describe('#utils calcChanges', () => {
...
+          should_be_changed: 'should_be_changed',
+          should_be_removed: 'should_be_removed'
+        }
+      }
+    ];
+    const assets = [
+      {
+        id: 'id3',
+        metadata: {
+          should_be_preserved: 'should_be_preserved',
```
turcottedanny, line 202: Maybe remove this one, if we want to test orphans metadata.

turcottedanny (commented):
Is there any other resources that we need to check for objectFields ?

test/tools/utils.tests.js
ggoodman: The test for orphans is the `should_be_removed` value in the `existing` array. Since it's not in the `assets` array, it  is simulating a property set in the dashboard but not reflected in the manifest.

src/tools/utils.js
ggoodman: Fixed in cb3cca4.

ggoodman:
> Is there any other resources that we need to check for objectFields ?

There is the possibility that we'd want to also do this for users but I haven't found exactly where we'd do that. I think that it's reasonable to defer that to a future PR.

src/tools/utils.js
```diff
@@ -62,7 +62,71 @@ export function dumpJSON(obj, spacing = 0) {
...
+          currentAssetState
+        )}`
+      );
+    }
+  }
+
+  return desiredAssetStateWithChanges;
+}
+
+export function calcChanges(assetType, assets, existing, identifiers = [ 'id', 'name' ], objectFields = [], allowDelete = false) {
```
turcottedanny, line 129: assetType still exists here. Also on line 180

test/tools/utils.tests.js
turcottedanny: `should_be_removed` got me confused, I thought it would be deleted, but it's intended to be remove from the diff.

src/tools/utils.js
ggoodman: Fixed in 392c52c.

turcottedanny (approved):
Good job!

PR #371, opened 2021-08-20
Bump js-yaml from 3.x to 4.x and move to kacl

ggoodman:
## Description

This PR bumps `js-yaml` from `3.x` to `4.x` in accordance with its [migration guide](https://github.com/nodeca/js-yaml/blob/master/migrate_v3_to_v4.md). This bump means that we're able to use the default _safe_ behaviour for both exports and imports. 

Notably, this means that we won't end up with values like `!<tag:yaml.org,2002:js/undefined> ''` that are not at all human friendly and were problematic when we used the `.safeLoad` functionality.

I believe that this is not a breaking change as any instances where deploy-cli produced incompatible output would not be valid as an input since we had been using `.dump()` but then using `.safeLoad()` which are not symmetrical. We could have used `.safeDump()` instead to address this. I think that the approach of adopting the latest, supported `js-yaml` version where all methods are the 'safe' equivalents by default is the better option.


ggoodman:
Note to reviewer: I will prepare a separate release PR with changelog items after this for the unreleased changes in master.

turcottedanny (approved):
LGTM!

PR #379, opened 2021-09-15
Allow set of AUTH0_AUDIENCE for custom domain

AliBazzi:
## ✏️ Changes

Separate `AUTH0_DOMAIN` value from audience value while retrieving an access token for management API.

## 🔗 References

We have a private appliance which doesn't expose default domains over the internet and only allows access to management API over customer facing domain (public facing).

Currently, the `auth0-cli` will set the audience value for Management API client based on the value of `AUTH0_DOMAIN`, that would work if the `AUTH0_DOMAIN` is being set as the default domain of the tenant, which is the audience as well for management API.

This PR open the possibility to set `AUTH0_AUDIENCE` value explicitly through config object and assign it to the `audience` of `clientCredentialsGrant ` or if not set then retain the current logic by setting the `audience` as `https://${config.AUTH0_DOMAIN}/api/v2/`

## 🎯 Testing

### DIFF
--- a/src/context/index.js
+++ b/src/context/index.js
@@ -43,7 +43,7 @@ export default async function(config) {
     });
 
     const clientCredentials = await authClient.clientCredentialsGrant({
-      audience: `https://${config.AUTH0_DOMAIN}/api/v2/`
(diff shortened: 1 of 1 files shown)


shushen (approved):
Thanks for the PR!

Issue #387, opened 2021-11-11
Unable to skip deploying organizations by not including organizations entry in the tenant yaml


mindbergh:
### Description

Similar to https://github.com/auth0/auth0-deploy-cli/issues/74, I expect the deployment to skip the organizations if the organizations entity is not present in the `tenant.yaml`. For example:

The following should trigger deletion of all organizations in the tenant if any.
```yaml
organization: []
```

The following should do nothing:
```yaml

```

However, when organizations entity is not present in yaml, the deployment still tries to fetch all organizations, and notifies the deletion can be done by enabling `AUTH0_ALLOW_DELETE`.

### Reproduction

`organizations` entry isn't present in `tenant.yaml`.

```
a0deploy import -c config.json -i tenant.yaml

...
2021-11-10T23:42:25.654Z - warn: Detected the following organizations should be deleted. Doing so may be destructive.
You can enable deletes by setting 'AUTH0_ALLOW_DELETE' to true in the config

{"id":"xxx","name":"xxx","display_name":"xxx","connections":[]}
...
```

#### Test

If organizations are added to [this test](https://github.com/auth0/auth0-deploy-cli/blob/master/test/context/yaml/context.test.js#L28) like other entities, the test fails.

### Environment

```
a0deploy --version
7.3.1
```

willvedd:
@mindbergh thanks for your patience and contribution! Your PR #388 solved this issue and has been added in our [7.5.0 release](https://github.com/auth0/auth0-deploy-cli/releases/tag/v7.5.0). At your convenience, please upgrade and let me know if you have any issues!

PR #388, opened 2021-11-11
Skip org deployment when org entity is not present in yaml

mindbergh:
## ✏️ Changes

When `organizations` isn't present in the yaml, returns `undefined` instead of `[]` so the deployment for organizations doesn't trigger.

## 🔗 References

#387 

## 🎯 Testing

### DIFF
--- a/src/context/yaml/handlers/organizations.js
+++ b/src/context/yaml/handlers/organizations.js
@@ -1,10 +1,8 @@
 async function parse(context) {
   const { organizations } = context.assets;
 
-  if (!organizations) return {};
-
   return {
-    organizations: organizations || []
+    organizations: organizations
   };
 }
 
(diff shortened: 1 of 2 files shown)


mindbergh:
Can someone from the maintainer team take a look?

mindbergh:
@shushen could you take a look?

marouanekadiri-sc:
This would fix https://github.com/auth0/auth0-deploy-cli/issues/402 for people who doesn't use `organisations` in their yaml file.

marouanekadiri-sc:
@mindbergh Is it possible to return an empty object in the `dump` function in the case if `organisations` is empty. To avoid the yaml file to have empty array after the `export`?

willvedd (approved):
Am able to confirm that this provides parity with how organizations are handled between the two formats. I've also tested that this works as described and allows folks to ignore organizations and be able to delete other types of resources.

willvedd:
@mindbergh Thank you for the contribution and sorry that this has taken so long to get merged in.

PR #391, opened 2021-11-19
Fall back to hook id when no name set

ggoodman:
## ✏️ Changes

Fixes an issue where a tenant having an old hook without the `hook-name` metadata property would be unable to use `deploy-cli`'s directory dump mode. It seems that between the API2 hooks facade and the CLI there's an impedence mismatch such that no `.name` property will come back. To work around this, we default the internal name representation to the hook's `.id`.

## 🔗 References

https://auth0team.atlassian.net/browse/DXEX-1940

## 🎯 Testing

### DIFF
--- a/src/tools/auth0/handlers/hooks.js
+++ b/src/tools/auth0/handlers/hooks.js
@@ -150,7 +150,7 @@ export default class HooksHandler extends DefaultHandler {
       // hooks.getAll does not return code and secrets, we have to fetch hooks one-by-one
       this.existing = await Promise.all(hooks.map((hook) => this.client.hooks.get({ id: hook.id })
         .then((hookWithCode) => this.client.hooks.getSecrets({ id: hook.id })
-          .then((secrets) => ({ ...hookWithCode, secrets })))));
+          .then((secrets) => ({ name: hook.id, ...hookWithCode, secrets })))));
 
       return this.existing;
     } catch (err) {


src/tools/auth0/handlers/hooks.js
```diff
@@ -150,7 +150,7 @@ export default class HooksHandler extends DefaultHandler {
       // hooks.getAll does not return code and secrets, we have to fetch hooks one-by-one
       this.existing = await Promise.all(hooks.map((hook) => this.client.hooks.get({ id: hook.id })
         .then((hookWithCode) => this.client.hooks.getSecrets({ id: hook.id })
-          .then((secrets) => ({ ...hookWithCode, secrets })))));
+          .then((secrets) => ({ name: hook.id, ...hookWithCode, secrets })))));
```
seejee, line 153: This LGTM, but should the proper fix be in API2 so that we always return a name on the hook result?

ggoodman: SGTM. After peeling back 20 (unnecessary?) layers of abstraction, the issue seems to be here: https://github.com/auth0/auth0-hooks-api/blob/0f801814243dd6e8b620900f1d6b4b4f42f2441b/packages/hooks-api-core/src/hooks/mappers/HookMapper.ts#L9-L17

ggoodman: https://github.com/auth0/auth0-hooks-api/pull/51

willvedd:
@ggoodman @seejee Hey you guys have any update on this? Looks like an associated PR might've been merged in? Would love to merge if possible.

shushen:
@willvedd This issue should be fixed in the backend - it's not a deploy-cli bug. I'll close this PR.

PR #405, opened 2022-02-01
[ESD-17871] bind client.connections to get scope with class methods

adamjmcgrath:
## ✏️ Changes

A change went in to `node-auth0@2.38.0` where we refactored the way we defined the various services methods, changing them from being defined using [wrapPropertyMethod](https://github.com/auth0/node-auth0/blob/fc3d2060af577d108c022d0863b072620e5b1829/src/utils.js#L50-L64) to regular class methods.

`wrapPropertyMethod` used to create getters and this code in `auth0-deploy-cli` relied on the method being defined as a getter rather than a class method and lost it's context with the latest version of node-auth0 https://github.com/auth0/auth0-deploy-cli/blob/6a7913d7409561878dc8c7dc7f6afa1e40efbccc/src/tools/auth0/handlers/databases.js#L52

Because the receiver in `Reflect.get(target, propertyKey, receiver)` is the value of `this` provided for the call to target if a getter is encountered (_not_ when a class method is encountered).

eg. 

```js
const oldAuth0 = Object.defineProperty(new function() {}, 'create', {
  get: function() { return () => console.log('this ->', this) }
})
const newAuth0 = new class { create() { console.log('this ->', this) } }

Reflect.get(oldAuth0, 'create', 'foo')()
// this -> 'foo'

Reflect.get(newAuth0, 'create', 'foo')()
// this -> undefined
```

Changing how we bind the method to [match the base getClientFN](https://github.com/auth0/auth0-deploy-cli/blob/99fe955c31e6d99c88c64a3992c610b9b7113395/src/tools/auth0/handlers/default.js#L44) fixes the issue.

Apologies for the disruption - we can't support this behaviour on `node-auth0` and we think it's unique enough to fix here.

## 🔗 References

Fixes https://auth0team.atlassian.net/browse/ESD-17871

## 🎯 Testing

Make sure you have a dependency on node-auth0 >= 2.38.0
Import a yml like the one described in [ESD-17871](https://auth0team.atlassian.net/browse/ESD-17871)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "ajv": "^6.12.6",
-        "auth0": "^2.37.0",
+        "auth0": "^2.39.0",
         "dot-prop": "^5.2.0",
         "fs-extra": "^7.0.0",
         "global-agent": "^2.1.12",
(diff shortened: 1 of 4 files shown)


willvedd (approved):
Looks good, thanks for looking into this.

willvedd:
This should be merged at some point in time, but recommending we rely on #406 as a more encompassing short-term fix as we take the time to understand the underlying causes more.

willvedd:
@adamjmcgrath Thanks for your help on this, it largely shaped #412 . However, I'm going to close since it's been made redundant.

PR #406, opened 2022-02-02
ESD-17871: Pinning Auth0 Node SDK to 2.37.0

willvedd:
## ✏️ Changes

Pinning Node SDK to version 2.37.0. Versions after that include a refactor to the implementation of resource class implementations that this repo depends on. This can be seen in [ESD-17871](https://auth0team.atlassian.net/browse/ESD-17871) where users are unable to create database connections with fresh installs that pull later version of the Node SDK. 

@adamjmcgrath has already submitted a fix with for this specific ESD ticket with #405 , but it is likely that other bugs have been introduced. So for that reason, I recommend pinning the version as a safer and more wholistic fix until we can better understand the scope.

**This is very much a temporary fix.** A more longterm solution needs to be investigated with @shushen and @TwelveNights  and @adamjmcgrath to better understand the underlying abstractions that are at the core of ESD-17871. 

## 🔗 References

- [ESD-17871](https://auth0team.atlassian.net/browse/ESD-17871) Main ESD ticket
- [ESD-17791](https://auth0team.atlassian.net/browse/ESD-17791) Related ESD ticket

## 🎯 Testing

Testing was performed manually by applying test configuration provided by customer and changing the versions of the dependency back and forth. 

If any automated tests can be applied for this scenario, please advise!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Fix errors caused by incompatibilities introduced by new versions of Auth0 SDK [#406]
 
 ## [7.3.5] - 2022-01-27
 ### Fixed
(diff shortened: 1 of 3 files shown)


shushen (approved):

shushen (changes requested):
@willvedd Could you please bump the version of the package and update the CHANGLOG?

shushen (approved):

PR #408, opened 2022-02-03
feat: expose error on actions

TwelveNights:
## ✏️ Changes

If there's an exception during the actions deploy stage, we do not expose the error. Instead, execution will continue (and perhaps fail in later steps). This pull request propagates any errors we may encounter in the deploy stage.

## 🔗 References

Work to support https://auth0team.atlassian.net/browse/ESD-17919

## 🎯 Testing


turcottedanny (approved):
LGTM

PR #412, opened 2022-02-15
ESD-17871: Binding function context of database methods after losing with SDK update

willvedd:
## ✏️ Changes

A change went in to `node-auth0@2.38.0` where we refactored the way we defined the various services methods, changing them from being defined using [wrapPropertyMethod](https://github.com/auth0/node-auth0/blob/fc3d2060af577d108c022d0863b072620e5b1829/src/utils.js#L50-L64) to regular class methods. It's important to note that this was not a regression on the `node-auth0` side, but rather changes in implementation details that the deploy CLI relied on.

The issue could be observed when updating the `node-auth` package to 2.38.0 or later and then adding a new database configuration. Then during import, the import would fail and log the following:

```shell
2022-02-15T20:10:18.462Z - error: Problem running command import during stage processChanges when processing type databases
2022-02-15T20:10:18.462Z - error: Cannot read properties of undefined (reading 'resource')
```

This change is largely based off [@adamjmcgrath](https://github.com/adamjmcgrath)'s #405 where he correctly binds the correct scope to the dynamically run class methods. **However**, this PR also goes a step further and removes the two major references to `Reflect.get` and replaces with more conventional bracket syntax. This change was introduced because the `Reflect.get` didn't serve a functional purpose and increased cognitive overhead when debugging the code. Example:

```js
Reflect.get( client, 'create')() // Current method
// versus

client.create() // More intuitive and functionally equivalent
```

## 🔗 References

[ESD-17871: a0deploy import fails when database is not created (node-auth0 2.38.0 regression)](https://auth0team.atlassian.net/browse/ESD-17871) 

## 🎯 Testing

**The bulk of this PR is test changes**. The important bit is enforcing that the `this` keyword is defined in all the important class methods like `create`, `update` and `delete` (I don't believe `get` and `getAll` necessarily require this change). The most important testing change is the in `test/tools/auth0/handlers/databases.tests.js` file (pointed out below), but I just went ahead and added these tests.

I may have gone a bit overboard with the testing changes, but I wanted to err on the paranoid side as I'm still learning the inner workings of this project.

### DIFF
--- a/src/tools/auth0/handlers/databases.js
+++ b/src/tools/auth0/handlers/databases.js
@@ -49,7 +49,7 @@ export default class DatabaseHandler extends DefaultHandler {
         });
     }
 
-    return Reflect.get(this.client.connections, fn, this.client.connections);
+    return this.client.connections[fn].bind(this.client.connections)
   }
 
   async getType() {
--- a/src/tools/auth0/handlers/databases.js
+++ b/src/tools/auth0/handlers/databases.js
@@ -49,7 +49,7 @@ export default class DatabaseHandler extends DefaultHandler {
         });
     }
 
-    return this.client.connections[fn].bind(this.client.connections)
+    return this.client.connections[fn].bind(this.client.connections);
   }
 
   async getType() {
--- a/src/tools/auth0/handlers/default.js
+++ b/src/tools/auth0/handlers/default.js
@@ -40,8 +40,8 @@ export default class DefaultHandler {
 
   getClientFN(fn) {
     if (typeof fn === 'string') {
-      const client = Reflect.get(this.client, this.type);
-      return Reflect.get(client, fn).bind(client);
+      const client = this.client[this.type]
+      return client[fn].bind(client);
     }
     return fn;
   }
--- a/src/tools/auth0/handlers/default.js
+++ b/src/tools/auth0/handlers/default.js
@@ -40,7 +40,7 @@ export default class DefaultHandler {
 
   getClientFN(fn) {
     if (typeof fn === 'string') {
-      const client = this.client[this.type]
+      const client = this.client[this.type];
       return client[fn].bind(client);
     }
     return fn;
--- a/test/tools/auth0/handlers/actions.tests.js
+++ b/test/tools/auth0/handlers/actions.tests.js
@@ -127,8 +127,8 @@ describe('#actions handler', () => {
             expect(params.id).to.equal(actionId);
             return Promise.resolve({ ...action, id: actionId });
           },
-          create: function(data){
-            expect(this).to.not.be.undefined
+          create: function(data) {
+            (() => expect(this).to.not.be.undefined)();
             expect(data).to.be.an('object');
             expect(data.name).to.equal('action-test');
             expect(data.supported_triggers[0].id).to.equal('post-login');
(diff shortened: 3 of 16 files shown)


test/tools/auth0/handlers/databases.tests.js
```diff
@@ -45,15 +45,16 @@ describe('#databases handler', () => {
...
     });
   });
 
   describe('#databases process', () => {
     it('should create database', async () => {
       const auth0 = {
         connections: {
-          create: (data) => {
+          create: function (data) {
+            expect(this).to.not.be.undefined
```
willvedd, line 57: This is the key test for preventing the precise issue causing ESD-17871.

test/tools/auth0/handlers/clients.tests.js
```diff
@@ -98,12 +99,14 @@ describe('#clients handler', () => {
     it('should update client', async () => {
       const auth0 = {
         clients: {
-          create: (data) => {
+          create: function(data){
+            expect(this).to.not.be.undefined
```
willvedd, line 103: Much of this PR is just asserting that the `this` keyword is defined in all the class methods.

src/tools/auth0/handlers/databases.js
```diff
@@ -49,7 +49,7 @@ export default class DatabaseHandler extends DefaultHandler {
         });
     }
 
-    return Reflect.get(this.client.connections, fn, this.client.connections);
+    return this.client.connections[fn].bind(this.client.connections)
```
willvedd, line 52: Binding the context of this function with `this.client.connections` is ultimately what fixes this issue. 

I make the extra step and swap `Reflect.get` with the array bracket syntax for clarity; especially after consulting with Adam, we deemed `Reflect` to be an obscure and confusing addition here.

src/tools/auth0/handlers/default.js
```diff
@@ -40,8 +40,8 @@ export default class DefaultHandler {
 
   getClientFN(fn) {
     if (typeof fn === 'string') {
-      const client = Reflect.get(this.client, this.type);
-      return Reflect.get(client, fn).bind(client);
+      const client = this.client[this.type]
+      return client[fn].bind(client);
```
willvedd, line 44: Similarly with [the core code fix](https://github.com/auth0/auth0-deploy-cli/pull/412/files#diff-e3ee2712757491d17ff1b85f517c4260ad50644cc9e76d31263652dec03190f0R52), I opt to swatch the syntax here from `Reflect.get` to the more straightforward object property bracket syntax.

test/tools/auth0/handlers/clientGrants.tests.js
```diff
@@ -34,7 +34,7 @@ describe('#clientGrants handler', () => {
       ];
 
       try {
-        await stageFn.apply(handler, [ { clientGrants: data } ]);
+        await stageFn.apply(handler, [{ clientGrants: data }]);
```
willvedd, line 37: Apologies with the formatting changes. This is something we'll have to standardize with linting.

test/tools/auth0/handlers/actions.tests.js
```diff
@@ -127,7 +127,8 @@ describe('#actions handler', () => {
             expect(params.id).to.equal(actionId);
             return Promise.resolve({ ...action, id: actionId });
           },
-          create: (data) => {
+          create: function(data) {
+            (() => expect(this).to.not.be.undefined)();
```
willvedd, line 131: Excuse the IIFE, it was the only way to quell the mass amount of `Expected an assignment or function call and instead saw an expression` linting errors.

sergiughf (approved):

adamjmcgrath (approved):
lgtm 👍

PR #414, opened 2022-02-15
Adding "dev" script to enable auto-build on file change

willvedd:
## ✏️ Changes

Introducing the `dev` script that can be run with `npm run dev` to build automatically on source code change. This reduces the manual toil of building after modifying source code. Although this could have been accomplished previously by running `npm run build -- watch`, that functionality wasn't obvious to me so I'm deciding to expose it more clearly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "pretest": "rimraf ./.nyc_output",
     "test": "cross-env NODE_ENV=test nyc mocha --recursive --require @babel/register test",
     "build": "rimraf ./lib && babel src -d lib",
+    "dev": "npm run build -- --watch",
     "prepare": "npm run build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -12,6 +12,7 @@
     "pretest": "rimraf ./.nyc_output",
     "test": "cross-env NODE_ENV=test nyc mocha --recursive --require @babel/register test",
     "build": "rimraf ./lib && babel src -d lib",
+    "dev": "npm run build -- --watch",
```
willvedd, line 15: I'd also be ok with naming this `build:dev` but `dev` is clear and a common convention I've followed in the past that encompasses all appropriate processes to develop the project.

sergiughf (approved):

Issue #373, opened 2021-08-23
Latest v7.2.1 outputting warnings about client_metadata


gparascandolo:
### Description

When trying to import tenant.yaml file with Auth0 Deploy CLI v7.2.1, we are encountering warnings about client_metadata for "undefined"

The clients still update after the warnings appear, as seen in the example logs below(with debug flag), but it is raising some concerns:
> 2021-08-23T21:15:21.278Z - warn: Detected that the client_metadata of the following undefined should be emptied. Doing so may be destructive.
> You can enable deletes by setting 'AUTH0_ALLOW_DELETE' to true in the config
> {"name":"Test Application","client_id":"siSPR......U1Dh"}
> 2021-08-23T21:15:21.280Z - debug: Start processChanges for clients [delete:0] [update:1], [create:0], [conflicts:0]
> 2021-08-23T21:15:21.624Z - debug: Stripping "Test Application" read-only fields ["jwt_configuration.secret_encoded","client_id"]
> 2021-08-23T21:15:23.531Z - info: Updated [clients]: {"name":"Test Application","client_id":"siSPR......U1Dh"}

The warning appears whether "AUTH0_ALLOW_DELETE" is set to true or false in the config file

Our config.json contains the following:

```
{
  "AUTH0_DOMAIN": "REDACTED.us.auth0.com",
  "AUTH0_CLIENT_ID": "SJZ......w7",
  "AUTH0_ALLOW_DELETE": true,
  "EXCLUDED_PROPS": {
    "clients": ["client_secret"],
    "connections": ["options.client_secret"]
  }
}
```

### Reproduction

1. a0deploy export config with YAML format  (`a0deploy export -f yaml -o output/to/directory/. --config_file path/to/config.json`)
2. a0deploy import the same exported tenant.yaml + config (`a0deploy import --input_file path/to/tenant.yaml --config_file path/to/config.json --debug`)

This warning is appearing consistently

mendhak:
This happens with the directory based import/export as well.  Our imports are polluted with such lines.  

> warn: Detected that the client_metadata of the following undefined should be emptied. Doing so may be destructive

Could someone from Auth0 explain what needs to be done here?  

[...]

PR #416, opened 2022-02-17
DXCDT-53: Preventing benign warnings when diff'ing object fields

willvedd:
## ✏️ Changes

As reported by #373, there was the following warning that was popping up when pushing clients to remote tenants:

```
2021-08-23T21:15:21.278Z - warn: Detected that the client_metadata of the following undefined should be emptied. Doing so may be destructive.
You can enable deletes by setting 'AUTH0_ALLOW_DELETE' to true in the config
{"name":"Test Application","client_id":"siSPR......U1Dh"}
2021-08-23T21:15:21.280Z - debug: Start processChanges for clients [delete:0] [update:1], [create:0], [conflicts:0]
2021-08-23T21:15:21.624Z - debug: Stripping "Test Application" read-only fields ["jwt_configuration.secret_encoded","client_id"]
2021-08-23T21:15:23.531Z - info: Updated [clients]: {"name":"Test Application","client_id":"siSPR......U1Dh"}
```

It's a warning, but a scary one at that. And worse, it's not actionable because contrary to the messaging, it would occur despite `AUTH0_ALLOW_DELETE` being set to true. I ultimately determined this warning to be benign, as in, it would never have caused any destructive changes. But it may have prevented some client metadata from being deleted properly.

The issue here was the logic in the `processChangedObjectFields` which would throw the warning about the `metadata` **object field**. An object field is an abstraction that seems to have been created to deal with key-value object properties, such as client metadata, and determining if these object properties should be deleted. Currently, the only place this is implemented is on client metadata, but in theory, it could be applied to any resource field that is a key-value object.

The fix here is to check to see if both the propose and current state of that object property are both empty, chalking it up as a no-op and then continuing to check the other object fields. The rest of the logic in the function is sound.

One issue that I discovered while working on this is that the `allowDelete` argument is never actually passed when calculating the diff, which is why this bug occurred regardless of what the `AUTH0_ALLOW_DELETE` config var was set to. This will need to be fixed in a following PR.

## 🎯 Testing

Exposed the `processChangedObjectFields` through an export and added unit tests to assert the expected behavior.

### DIFF
--- a/src/tools/utils.js
+++ b/src/tools/utils.js
@@ -76,6 +76,13 @@ function processChangedObjectFields(handler, desiredAssetState, currentAssetStat
 
   // eslint-disable-next-line no-restricted-syntax
   for (const fieldName of objectFields) {
+    const areDesiredStateAndCurrentStateEmpty = Object.keys(desiredAssetState[fieldName] || {}).length === 0 && Object.keys(currentAssetState[fieldName] || {}).length === 0;
+    if (areDesiredStateAndCurrentStateEmpty) {
+      // If both the desired state and current state for a given object is empty, it is a no-op and can skip
+      // eslint-disable-next-line no-continue
+      continue;
+    }
+
     // A desired state that omits the objectField OR that has it as an empty object should
     // signal that all fields should be removed (subject to ALLOW_DELETE).
     if (desiredAssetState[fieldName] && Object.keys(desiredAssetState[fieldName]).length) {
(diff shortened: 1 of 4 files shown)


src/tools/utils.js
```diff
@@ -71,11 +71,20 @@ export function dumpJSON(obj, spacing = 0) {
  * @param {boolean} [allowDelete=false]
  * @returns T
  */
-function processChangedObjectFields(handler, desiredAssetState, currentAssetState, objectFields = [], allowDelete = false) {
+export function processChangedObjectFields({
+  handler, desiredAssetState, currentAssetState, allowDelete = false
+}) {
```
willvedd, line 76: Changing the function to accept an object of parameters instead of several arguments. Also, I removed the `objectFields` argument because it can be derived from the handler directly.

src/tools/utils.js
```diff
@@ -71,11 +71,20 @@ export function dumpJSON(obj, spacing = 0) {
...
 
   // eslint-disable-next-line no-restricted-syntax
-  for (const fieldName of objectFields) {
+  for (const fieldName of handler.objectFields) {
+    const areDesiredStateAndCurrentStateEmpty = Object.keys(desiredAssetState[fieldName] || {}).length === 0 && Object.keys(currentAssetState[fieldName] || {}).length === 0;
+    if (areDesiredStateAndCurrentStateEmpty) {
+      // If both the desired state and current state for a given object is empty, it is a no-op and can skip
+      // eslint-disable-next-line no-continue
+      continue;
+    }
```
willvedd, line 86: This is ultimately what will quell the warnings originally proposed in #373 .

src/tools/utils.js
```diff
@@ -88,15 +97,17 @@ function processChangedObjectFields(handler, desiredAssetState, currentAssetStat
...
             if (allowDelete) {
               desiredAssetStateWithChanges[fieldName][
                 currentObjectFieldPropertyName
               ] = null;
             } else {
+              // If deletes aren't allowed, do outright delete the property within the object
               log.warn(
-                `Detected that the ${fieldName} of the following ${
-                  handler.name
+                `Detected that the ${fieldName} of the following ${handler.name || handler.id || ''
```
willvedd, line 110: Added the ID, although it might not be super helpful in some cases, but it does prevent this from printing out `undefined` in the logs.

sergiughf (approved):

Issue #415, opened 2022-02-17
Error importing client with keyword replace mappings


JHOmega:
**Please do not report security vulnerabilities here**. The [Responsible Disclosure Program](https://auth0.com/whitehat) details the procedure for disclosing security issues.

**Thank you in advance for helping us to improve this library!** Please read through the template below and answer all relevant questions. Your additional work here is greatly appreciated and will help us respond as quickly as possible. For general support or usage questions, use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com/). Finally, to avoid duplicates, please search existing Issues before submitting one here.

By submitting an Issue to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description
First of all i want make clear that i want to use the directory structure. 
I'm trying to setup the auth0 deploy cli to deploy to multiple environments. For one environment i want a list of callbacks configured in my client.json and for another environment a different list with a different number of callbacks.

### Reproduction

my dev.json:
```
{
    "AUTH0_DOMAIN": "obscured.xx.auth0.com",
    "AUTH0_CLIENT_ID": "obscured",
    "EXCLUDED_PROPS": {
        "clients": ["client_secret"],
        "connections": ["options.client_secret"]
      },
    "AUTH0_KEYWORD_REPLACE_MAPPINGS": { 
        "AUTH0_TENANT_NAME": "obscured",
        "API_AUDIENCE": "some-audience"
        "ALLOWED_CALLBACKS": ["http://somewhere-obscured.com/callback", http://localhost:xxxx/callback"]
        "ALLOWED_LOGOUT_URLS": ["http://somewhere-obscured.com/", "http://localhost:xxxx/"]
    },
    "AUTH0_ALLOW_DELETE": true
}
```

my client config: (My App.json)
```
{
...
"name": "My App",
  "description": "The application for obscured",
  "callbacks": @@ALLOWED_CALLBACKS@@,
  "allowed_logout_urls": @@ALLOWED_LOGOUT_URLS@@,
....
}
```

and i get: 
```
Error parsing JSON from metadata file: clients/My App.json, because: Unexpected token @ in JSON at position 140
```

I have tried by enclosing @@ in quotes: "@@ALLOWED_CALLBACKS@@" the error i get then is:
```
[...]

Issue #184, opened 2019-11-22
Keyword replace mappings (# / @) in configs don't support concatenating arrays


gaui:
### Description

> Provide a clear and concise description of the issue, including what you expected to happen.

Placeholders `##x##` or `@@x@@` don't support concatenating arrays. Sometimes you would like to concatenate arrays, e.g. allowed callback URLs, logout URLs or web origins. For example having base/global allowed URLs defined, and then adding to them for all clients/applications.

**config.json**

```json
{
	"AUTH0_KEYWORD_REPLACE_MAPPINGS": {
		  "GLOBAL_WEB_ORIGINS": [
			"http://local.me:8080",
			"http://localhost",
			"http://localhost:3000"
		  ]
	}
}
```

**def/clients/A.json**

Either `##`:

```json
{
    "web_origins": [
        "##GLOBAL_WEB_ORIGINS##",
        "http://a.foo.com",
        "https://a.foo.com"
    ]
}
```

or `@@`

```json
{
    "web_origins": [
        @@GLOBAL_WEB_ORIGINS@@,
        "http://a.foo.com",
        "https://a.foo.com"
    ]
}
```

**Expected result:**

```json
{
    "web_origins": [
        "http://local.me:8080",
        "http://localhost",
        "http://localhost:3000",
        "http://a.foo.com",
        "https://a.foo.com"
    ]
}
```

**Actual result:**

I get an error for `@@` but for `##` I get the following:

```json
{
    "web_origins": [
        "http://local.me:8080,http://localhost,http://localhost:3000",
        "http://a.foo.com",
        "https://a.foo.com"
    ]
}
```

### Environment

> Please provide the following:

- **Version of this library used:** 3.6.3
- **Other relevant versions (language, server software, OS, browser):** Node 8.15.1, Linux Mint,  Ubuntu 18.04
- **Other modules/plugins/libraries that might be involved:** N/A

aaron-fuleki:
Same. I've tried various incantations of escaping/nesting, and none have worked. I can fix things manually in the web UI, but the CI pipeline will pave over that on the next deploy, so that's not a valid workaround.

jermeo:
Hello, same problem for me.
Currently, we use several reserved spaces for temporary solution but it's not flexible, like this:
{
    "web_origins": [
        ##GLOBAL_WEB_ORIGIN_1##,
        ##GLOBAL_WEB_ORIGIN_2##,
        "http://a.foo.com",
[...]

PR #421, opened 2022-02-24
DXCDT-56: Fixing @@ array replacement to work with and without wrapped quotes

willvedd:
## ✏️ Changes

The way that the `@@FOO@@` array variable replacement worked has caused a bit of confusion from users. Firstly, it's not obvious if these markers need to be wrapped in quotes or not. Without them, it invalidates the JSON file that they're used in, but if they do use quotes, the replacement won't work as expected. This can be seen in #415 and #184. 

For example, as it currently is implemented, developers would need to have the following configurations:
```json
{
  "web_origins": @@WEB_ORIGINS@@
}
```
This is _invalid_ JSON and a bit of a wonky thing to force onto the developer. With this change, they can now use the array replacements with or without wrapped quotes:

```json
{
   "grant_types": @@GRANT_TYPES@@,
   "web_origins": "@@WEB_ORIGINS@@"
}
 
```

The tricky part here is that this needs to work both for JSON and YAML input files. This fix is primarily for JSON configurations to maintain the validity of the file. However, YAML is a bit less strict about having the @@ array markers wrapped in quotes, however that is still supported, albeit not with a desired outcome. Examples of this can be seen in the test YAML test cases below.

A big bulk of this PR is adding tests to confirm and enforce the behavior that exists currently.

## 🎯 Testing

Separated out `keywordArrayReplace` and `keywordStringReplace` into their own functions so that they could have their own unit tests applied. Also added higher-level unit tests to the `keywordReplace` function to confirm expected results for both YAML and JSON inputs.

✅This change has unit test coverage

### DIFF
--- a/src/tools/utils.js
+++ b/src/tools/utils.js
@@ -4,18 +4,32 @@ import dotProp from 'dot-prop';
 import _ from 'lodash';
 import log from './logger';
 
+export function keywordArrayReplace(input, mappings) {
+  Object.keys(mappings).forEach(function(key) {
+    // Matching against two sets of patterns because a developer may provide their array replacement keyword with or without wrapping quotes. It is not obvious to the developer which to do depending if they're operating in YAML or JSON.
+    const pattern = `@@${key}@@`;
+    const patternWithQuotes = `"${pattern}"`;
+
+    const regex = new RegExp(`${patternWithQuotes}|${pattern}`, 'g');
+    input = input.replace(regex, JSON.stringify(mappings[key]));
+  });
+  return input;
+}
+
+export function keywordStringReplace(input, mappings) {
+  Object.keys(mappings).forEach(function(key) {
+    const regex = new RegExp(`##${key}##`, 'g');
+    input = input.replace(regex, mappings[key]);
+  });
+  return input;
+}
+
 export function keywordReplace(input, mappings) {
   // Replace keywords with mappings within input.
   if (mappings && Object.keys(mappings).length > 0) {
-    Object.keys(mappings).forEach(function(key) {
-      const re = new RegExp(`##${key}##`, 'g');
-      input = input.replace(re, mappings[key]);
-    });
+    input = keywordStringReplace(input, mappings);
 
-    Object.keys(mappings).forEach(function(key) {
-      const re = new RegExp(`@@${key}@@`, 'g');
-      input = input.replace(re, JSON.stringify(mappings[key]));
-    });
+    input = keywordArrayReplace(input, mappings);
   }
   return input;
 }
--- a/src/tools/utils.js
+++ b/src/tools/utils.js
@@ -53,7 +53,7 @@ export function convertClientNamesToIds(names, clients) {
   return [ ...unresolved, ...result ];
 }
 
-export function loadFile(file, mappings) {
+export function loadFileAndReplaceKeywords(file, mappings) {
   // Load file and replace keyword mappings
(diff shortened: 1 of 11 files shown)


src/tools/utils.js
```diff
@@ -4,18 +4,32 @@ import dotProp from 'dot-prop';
 import _ from 'lodash';
 import log from './logger';
 
+export function keywordArrayReplace(input, mappings) {
```
willvedd, line 7: Separated out and exported to enable unit testing.

src/tools/utils.js
```diff
@@ -4,18 +4,32 @@ import dotProp from 'dot-prop';
...
+  Object.keys(mappings).forEach(function(key) {
+    // Matching against two sets of patterns because a developer may provide their array replacement keyword with or without wrapping quotes. It is not obvious to the developer which to do depending if they're operating in YAML or JSON.
+    const pattern = `@@${key}@@`;
+    const patternWithQuotes = `"${pattern}"`;
```
willvedd, line 11: The key functional change in this PR is using regex to replace the `@@` pattern with and without wrapping quotes.

src/tools/utils.js
```diff
@@ -4,18 +4,32 @@ import dotProp from 'dot-prop';
...
+
+    const regex = new RegExp(`${patternWithQuotes}|${pattern}`, 'g');
+    input = input.replace(regex, JSON.stringify(mappings[key]));
```
willvedd, line 14: The key difference between this function and the below `keywordStringReplace` function is the `JSON.stringify` call.

test/tools/utils.tests.js
```diff
@@ -439,3 +439,98 @@ describe('#utils processChangedObjectFields', () => {
...
+      STRING_REPLACEMENT: 'baz',
+      OTHER_REPLACEMENT: 'lol'
+    };
+
+    const inputYAML = `
+    ---
+    stringReplaceNoQuotes: ##STRING_REPLACEMENT##
+    stringReplaceWithQuotes: "##STRING_REPLACEMENT##"
+    arrayReplace: @@ARRAY_REPLACEMENT@@
+    arrayReplaceWithQuotes: "@@ARRAY_REPLACEMENT@@"
```
willvedd, line 470: Unlikely that someone would wrap this value in quotes when using YAML, but at least it's technically supported. I don't believe this behavior will interfere with any current customer workflows.

test/tools/utils.tests.js
```diff
@@ -439,3 +439,98 @@ describe('#utils processChangedObjectFields', () => {
...
+  it('should replace string keywords and array keywords in a JSON file', () => {
+    const mapping = {
+      ARRAY_REPLACEMENT: [ 'foo', 'bar' ],
+      STRING_REPLACEMENT: 'baz',
+      OTHER_REPLACEMENT: 'lol'
+    };
+    const inputJSON = '{ "arrayReplaceNoQuotes": @@ARRAY_REPLACEMENT@@, "arrayReplaceWithQuotes": "@@ARRAY_REPLACEMENT@@", "stringReplace": "##STRING_REPLACEMENT##", "noReplace": "OTHER_REPLACEMENT" }';
+    const output = utils.keywordReplace(inputJSON, mapping);
+
+    expect(() => JSON.parse(output)).to.not.throw();
```
willvedd, line 453: This is the big test, just making sure the output JSON is validated.

sergiughf (approved):
LGTM 👍🏻

PR #422, opened 2022-02-24
Distinguishing between two loadFile functions

willvedd:
## ✏️ Changes

There are currently two `loadFile` functions, one is a utility function used to load and replace keywords. The other is a method on the resource handler class that allows extendable functionality to how configuration files are loaded, ultimately calling the former.

Both functions do similar things and one calls the other, but it is confusing when developing to distinguish between the two. This PR attempts to reduce the confusion by naming the base level utility function to `loadFileAndReplaceKeywords`. This will make reasoning about the keyword replacement code a bit more reasonable.

### DIFF
--- a/src/tools/utils.js
+++ b/src/tools/utils.js
@@ -53,7 +53,7 @@ export function convertClientNamesToIds(names, clients) {
   return [ ...unresolved, ...result ];
 }
 
-export function loadFile(file, mappings) {
+export function loadFileAndReplaceKeywords(file, mappings) {
   // Load file and replace keyword mappings
   const f = path.resolve(file);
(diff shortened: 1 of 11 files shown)


src/tools/utils.js
```diff
@@ -53,7 +53,7 @@ export function convertClientNamesToIds(names, clients) {
   return [ ...unresolved, ...result ];
 }
 
-export function loadFile(file, mappings) {
+export function loadFileAndReplaceKeywords(file, mappings) {
```
willvedd, line 56: The actual function declaration.

sergiughf (approved):

Issue #420, opened 2022-02-23
Inconsistent connection update behavior, potentially causing client_secret deletion


lundybernard:
### Description

deploying config `a0deploy deploy -c tenant/config.json --input_file tenant/tenant.yaml`

results in error:
> error: Problem running command deploy during stage processChanges when processing type connections
> error: Problem updating connections {"name":"###","id":"###"}
> Bad Request: Before creating Windows Azure Active Directory connections you must configure the client_id and client_secret settings in the dashboard. For more information go to https://docs.auth0.com/waad-clientid

also:
> error: Problem running command deploy during stage processChanges when processing type connections
> error: Problem updating connections {"name":"####","id":"####"}
> Bad Request: The signing certificate is not valid. Please provide a .pem or .cer certificate base64 encoded public key

### Reproduction
1. Export an existing environment, including one or more AzureAD connections

We are stripping out some sensitive information on export

config.json:
```
{
...
  "AUTH0_ALLOW_DELETE": false,
  "AUTH0_EXPORT_IDENTIFIERS": false,
  "AUTH0_EXCLUDED_RULES": [ "rule-1-name" ],
  "AUTH0_EXCLUDED_CONNECTIONS" : [
    "####",
    "smoke-test"
  ],
  "EXCLUDED_PROPS": {
    "connections": [
        "options.client_secret",
        "options.admin_access_token",
        "options.admin_access_token_expiresin",
        "options.cert",
        "options.signingCert"
    ]
  },
    "INCLUDED_PROPS": {
      "rulesConfigs": [ "key" ]
  }
}
```
tenant.yaml:
```
connections:
  - name: ####
    strategy: waad
    enabled_clients:
      - ####
    is_domain_connection: false
    options:
      domain: ####.com
      granted: true
      tenantId: ####
      client_id: ####
      use_wsfed: false
      app_domain: ####.auth0.com
      ext_groups: true
      ext_profile: true
      thumbprints:
        - ####
        - ####
        - ####
        - ####
        - ####
        - ####
      identity_api: azure-active-directory-v1.0
      basic_profile: false
      tenant_domain: ####.com
      waad_protocol: openid-connect
      domain_aliases:
        - ####.com
      api_enable_users: false
      ext_nested_groups: true
      useCommonEndpoint: false
      should_trust_email_verified_connection: ####
    show_as_button: false
```

Deploy the config: `a0deploy deploy -c tenant/config.json --input_file tenant/tenant.yaml`

[...]

PR #430, opened 2022-03-04
DXCDT-66: Add excluded properties into connections update payload

willvedd:
### Problem
Developers [reporting of issues](https://github.com/auth0/auth0-deploy-cli/issues/420) where properties excluded through the `EXCLUDED_PROPS` configuration values were properly excluding them on export, but on subsequent imports were effectively being deleted. This occurs because of a small quirk with the [PATCH connections endpoint](https://auth0.com/docs/api/management/v2#!/Connections/patch_connections_by_id):

>Note: if you use the options parameter, the whole options object will be overridden, so ensure that all parameters are present

Depending on the connection type, the result of this API call with critical excluded properties absent results in an error or success. In both cases, it precludes the ability to manage connections with excluded values. Fortunately as far as our team can tell, this _only_ affects connections. 

The solution here is to pluck the values of excluded properties from existing connections and add into the PATCH payload so that the excluded values are preserved in the tenant. The hope here is that the blast radius will be quite small and given the amount of test cases applied, should be fairly safe to introduce; we don't believe it's wise to apply a wider solution at this time.

## 🔗 References

- Relevant Github issue: #420 

## 🎯 Testing

Added the following test cases to the newly introduced `addExcludedConnectionPropertiesToChanges` function:

- Adding excluded properties when absent in the proposed changes
- Adding entire `options` property when excluded
- No change to payload if no config exists
- No change to payload if no relevant excluded properties exist
- No change to payload if no proposed updates exist

### DIFF
--- a/src/tools/auth0/handlers/connections.js
+++ b/src/tools/auth0/handlers/connections.js
@@ -1,7 +1,7 @@
 import dotProp from 'dot-prop';
+import _ from 'lodash';
 import DefaultHandler, { order } from './default';
 import { filterExcluded, convertClientNameToId, getEnabledClients } from '../../utils';
-import _ from 'lodash'
 
 export const schema = {
   type: 'array',
@@ -15,7 +15,7 @@ export const schema = {
       realms: { type: 'array', items: { type: 'string' } },
       metadata: { type: 'object' }
     },
-    required: ['name', 'strategy']
+    required: [ 'name', 'strategy' ]
   }
 };
 
@@ -37,7 +37,7 @@ export const addExcludedConnectionPropertiesToChanges = ({
   const excludedOptions = excludedFields.filter(
     // Only include fields that pertain to options
     (excludedField) => excludedField.startsWith('options')
-  )
+  );
 
   const newProposedUpdates = proposedChanges.update.map((proposedConnection) => {
     const currConnection = existingConnectionsMap[proposedConnection.id];
@@ -70,7 +70,7 @@ export default class ConnectionsHandler extends DefaultHandler {
     super({
       ...config,
       type: 'connections',
-      stripUpdateFields: ['strategy', 'name']
+      stripUpdateFields: [ 'strategy', 'name' ]
     });
   }
 
--- a/src/tools/auth0/handlers/connections.js
+++ b/src/tools/auth0/handlers/connections.js
@@ -1,6 +1,7 @@
 import dotProp from 'dot-prop';
 import DefaultHandler, { order } from './default';
 import { filterExcluded, convertClientNameToId, getEnabledClients } from '../../utils';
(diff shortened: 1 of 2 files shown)


test/tools/auth0/handlers/connections.tests.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable consistent-return */
```
willvedd, line 1: Just having some issue with the strictness of the linter, couldn't find a easier way to refactor.

src/tools/auth0/handlers/connections.js
```diff
@@ -17,6 +18,50 @@ export const schema = {
...
+  existingConnections,
+  config
+}) => {
+  if (proposedChanges.update.length === 0) return proposedChanges;
+
+  const excludedFields = config()?.EXCLUDED_PROPS?.connections || [];
+  if (excludedFields.length === 0) return proposedChanges;
+
+  const newProposedUpdates = proposedChanges.update.map((proposedConnection) => {
+    const currConnection = existingConnections.find(({ id }) => id === proposedConnection.id);
```
adamjmcgrath, line 36: Couple of loops you can do outside of `proposedChanges.update` for efficiency

```js
const existingConnectionsMap = _.keyBy(existingConnections, 'id');
const excludedOptions = excludedFields.filter(
  (excludedField) => excludedField.startsWith('options')
)

const newProposedUpdates = proposedChanges.update.map((proposedConnection) => {
  const currConnection = existingConnectionsMap[proposedConnection.id];
  const currentExcludedPropertyValues = excludedOptions.reduce((agg, excludedField) => {
    ...
```

src/tools/auth0/handlers/connections.js
```diff
@@ -17,6 +18,50 @@ export const schema = {
...
+
+    if (currConnection === undefined) return proposedConnection;
+
+    // Only include fields that pertain to options
+    const currentExcludedPropertyValues = excludedFields.filter((excludedField) => excludedField.indexOf('options') === 0).reduce((agg, excludedField) => {
+      if (!dotProp.has(currConnection, excludedField)) return agg;
```
adamjmcgrath, line 42: Should you also check `dotProp.has(agg, excludedField);`?

Is there potentially a valid use case for `proposedChanges` to include excluded fields?

willvedd: I think the `dotProp.has(agg, excludedField);` would be checking for the case of duplicated excluded fields, which isn't a valid use case.

>Is there potentially a valid use case for proposedChanges to include excluded fields?

Not to my knowledge, otherwise folks would just unexclude to incorporate those changes.

sergiughf (approved):

adamjmcgrath (approved):

Issue #172, opened 2019-10-11
`--no-env` not honored


tredmon:
**Please do not report security vulnerabilities here**. The [Responsible Disclosure Program](https://auth0.com/whitehat) details the procedure for disclosing security issues.

**Thank you in advance for helping us to improve this library!** Please read through the template below and answer all relevant questions. Your additional work here is greatly appreciated and will help us respond as quickly as possible. For general support or usage questions, use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com/). Finally, to avoid duplicates, please search existing Issues before submitting one here.

By submitting an Issue to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

> Provide a clear and concise description of the issue, including what you expected to happen.

The `--no-env` option is not honored. I would expect that when `--no-env` is passed environment variables would be ignored.

### Reproduction

> Detail the steps taken to reproduce this error, what was expected, and whether this issue can be reproduced consistently or if it is intermittent.
>
> Where applicable, please include:
>
> - Code sample to reproduce the issue
> - Log files (redact/remove sensitive information)
> - Application settings (redact/remove sensitive information)
> - Screenshots
```
$ AUTH0_DOMAIN=some-valid-domain.auth0.com AUTH0_CLIENT_ID=some-valid-client-id AUTH0_CLIENT_SECRET=some-valid-client-secret a0deploy import -i config/dir --no-env
2019-10-11T20:36:49.064Z - info: Processing directory config
2019-10-11T20:36:49.067Z - info: Getting access token for some-valid-client-id/some-valid-domain.auth0.com
2019-10-11T20:36:51.109Z - info: Import Successful
```
Note the "Import Successful" despite passing `--no-env` and not having passed a `-c config.json` containing a domain, client_id, or client_secret.

### Environment

> Please provide the following:

- **Version of this library used:** 3.6.1
- **Version of the platform or framework used, if applicable:**
- **Other relevant versions (language, server software, OS, browser):**
- **Other modules/plugins/libraries that might be involved:**

rcauquil:
any news on that, I have the same issue and have to make an uggly workaround

willvedd:
[...]

PR #432, opened 2022-03-08
DXCDT-47: Adding ability to prevent environment variable inheritance

willvedd:
## ✏️ Changes

As pointed out in #172 , there is currently [a documented `--no-env` flag](https://auth0.com/docs/deploy-monitor/deploy-cli-tool/environment-variables-and-keyword-mappings), but it isn't actually implemented in the codebase.  Excerpt from the docs:

>By default, the tool also merges your current environment variables and overrides AUTH0_KEYWORD_REPLACE_MAPPINGS that have the same top key. You can disable this via the command line with the --no-env option.

This PR leverages the undocumented, but existing `--env` flag to disable that env var inheritance by running `--env=false`. A documentation follow-up will need to be done to correct, but they are functionally equivalent.

Wether or not the default behavior of this inheritance is a good idea is a discussion for another time. At the very least, this can be circumvented.

## 🔗 References

- [Deploy CLI Tool Environment Variables and Keyword Mappings](https://auth0.com/docs/deploy-monitor/deploy-cli-tool/environment-variables-and-keyword-mappings)
- #172 

## 🎯 Testing

This PR doesn't have tests because the new introductions of `.ts` source files has broken our current testing and linting system. A follow-up PR is necessary to fix these processes so that tests can be added to this feature.

Because of this, this PR can and should not be merged in!


src/args.ts
```diff
@@ -29,8 +29,7 @@ function getParams() {
       },
       env: {
         alias: 'e',
-        describe: 'Override the mappings in config with process.env environment variables.',
-        type: 'string',
```
willvedd, line 33: `type: 'string'` is contradictory to `boolean: true` so went ahead and removed it.

src/args.ts
```diff
@@ -64,6 +63,12 @@ function getParams() {
         describe: 'The client secret, this allows you to encrypt the secret in your build configuration instead of storing it in a config file',
         type: 'string'
       },
+      env: {
+        alias: 'e',
+        describe: 'Override the mappings in config with environment variables.',
+        boolean: true,
+        default: false
+      },
```
willvedd, line 71: Added the `env` flag for the import script for parity, however, this time defaulting to false so the behavior remains the same.

src/commands/export.ts
```diff
@@ -6,23 +6,37 @@ import log from '../logger';
 import { isDirectory } from '../utils';
 import setupContext from '../context';
 
-export default async function deploy(params) {
```
willvedd, line 9: This function being named `deploy` is a clear mistake, went ahead and changed it to `exportCMD`. Cannot be `export` because it's a reserved keyword.

src/commands/export.ts
```diff
@@ -6,23 +6,37 @@ import log from '../logger';
...
     export_ids: exportIds,
-    secret
+    secret: clientSecret,
+    env: shouldInheritEnv = false,
   } = params;
 
-  nconf.env().use('memory');
+  if(shouldInheritEnv){
+    nconf.env().use('memory');
+  }
```
willvedd, line 33: This is the bulk of this PR.

src/commands/import.ts
```diff
@@ -4,17 +4,33 @@ import { deploy as toolsDeploy } from '../tools';
...
     config: configObj,
-    env,
-    secret
+    env: shouldInheritEnv = false,
+    secret: clientSecret
   } = params;
 
-  nconf.env().use('memory');
+  if(shouldInheritEnv){
+    nconf.env().use('memory');
```
willvedd, line 29: This is the bulk of this PR.

adamjmcgrath:
@willvedd 

> Because of this, this PR can and should not be merged in!

Do you mean that this should not be merged? If so - you can create a Draft PR

sergiughf (approved):

willvedd:
Having really a heck of time getting unit tests together for this. For one, there are no existing tests for the import and export command functions. But what's trickier, is that the underlying configuration storage dependency, `nconf` doesn't seem to have a functioning reset method, or at least I can't find any working examples.. Which means that once environment variables get loaded in and stored via `nconf` they cannot be cleared between unit tests, creating an issue with separation.

I'm opting to skip tests for this feature. I've tested it fairly thoroughly and have individual tests for both the import and export functions, but they cannot run together in the same test suite. I've just been spinning my wheels on this for a while. It's not a great precedent to set, I know, but for the sake of moving forward, I'm going create a jira ticket and move on.

johnwiseheart:
As far as I can tell, this is a breaking change, as previously using environment variables (eg `AUTH0_CLIENT_SECRET=xyz a0deploy ...`) worked with no `--env` flag, and now one is required. Am I understanding this wrong?

Issue #276, opened 2020-09-08
Universal Login maintenance


nason:
**Please do not report security vulnerabilities here**. The [Responsible Disclosure Program](https://auth0.com/whitehat) details the procedure for disclosing security issues.

**Thank you in advance for helping us to improve this library!** Your attention to detail here is greatly appreciated and will help us respond as quickly as possible. For general support or usage questions, use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com/). Finally, to avoid duplicates, please search existing Issues before submitting one here.

By submitting an Issue to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

I'd like to be able to manage my tenant's universal login page template with this CLI

https://auth0.com/docs/universal-login/new-experience/universal-login-page-templates

### Describe the ideal solution

The CLI imports and exports universal login page templates, and possibly other branding configuration

## Alternatives and current work-arounds

Currently, the only way we can manage these settings is manually via API.

We're considering scripting some kind of sync but it would be much better if this was managed by this CLI with the rest of our config

tyroneerasmus:
Just ran into this limitation too.

Anyone have any cobbled together scripts for doing this they can share? Otherwise I guess I am coding this for our CI/CD process and making it sit alongside our invocations of `a0deploy`

SAGV:
We are also hitting the limitation :-(

willvedd:
Folks, thanks for the feedback around this. It's definitely a reasonable request to manage all aspects of the universal login experience. There are a few nuanced technical reasons why I suspect this hasn't been implemented up to this point, but I'll spare you the explanation. But rest assured, we're looking closely at this project and its deficiencies, including this specific issue.

willvedd:
As noted in #315, the favicon URL should be a supported configuration to include under this general issue.

willvedd:
Good news folks, this has been added in the [7.6.0](https://github.com/auth0/auth0-deploy-cli/releases/tag/v7.6.0) release! Please upgrade and give it a go. If any issues arise, feel empowered to make an issue.

zebapy:
it's annoying that it requires a customized domain to pull settings though I can modify the settings in the UI.

PR #438, opened 2022-03-09
New branding template support

toddbaert:
:warning: work in progress, I've yet to add the yaml handler, doing that tomorrow, but I figured it wouldn't be bad to have my choices so far looked at :warning: 

## ✏️ Changes

- adds support for new universal login templates.
- only handles templates if custom domains are activated (custom templates are not allowed unless tenant is using a custom domain). For this reason, the `read:custom_domains` scope must be added to the client the CLI runs as.
- tried to make the code somewhat future proof for additional templates, though only "universal login" exists at the moment.

## 🔗 References

https://github.com/auth0/auth0-deploy-cli/issues/276

## 🎯 Testing

With a tenant that has a custom domain configured, use https://auth0.com/docs/api/management/v2#!/Branding/put_universal_login, and add a template for universal login. Run the `export` command, and find the template at `./branding/templates/universal-login.html`. Change the file, and run the import. Use  https://auth0.com/docs/api/management/v2#!/Branding/get_universal_login and see the template is updated.


src/tools/auth0/handlers/brandingTemplates.js
```diff
@@ -0,0 +1,61 @@
...
+    if (!this.client.branding || typeof this.client.branding.getUniversalLoginTemplate !== 'function') {
+      return {};
+    }
+
+    // in case client version does not custom domains
+    if (!this.client.customDomains || typeof this.client.customDomains.getAll !== 'function') {
+      return {};
+    }
+
+    // if there's no custom domains, skip, feature requires custom domain.
```
toddbaert, line 34: Because custom domains are required for this feature, I think it makes sense to bail out early if there's none configured.

test/tools/auth0/handlers/branding.tests.js
```diff
@@ -19,22 +19,30 @@ describe('#branding handler', () => {
...
-    it('should update branding settings', async () => {
+    it('should update branding settings, removing template property', (done) => {
       const auth0 = {
         branding: {
           updateSettings: (params, data) => {
             expect(data).to.be.an('object');
+            expect(data.templates).to.equal(undefined);
             expect(data.logo_url).to.equal('https://example.com/logo.png');
-            return Promise.resolve(data);
+            done();
```
toddbaert, line 29: The previous version of this test was fundamentally broken. The expectations were all in the stubbed `updateSettings()` function, this means that if that function call is completely removed from the production code, the test would run fine without any expectations ever being run at all.

To fix this, I've removed the `async` keyword, and instead added a `done()` callback to the stubbed function. This means that the test will fail (timeout) if the expected stub is not called.

willvedd (commented):
@toddbaert Appreciate you submitting this! Looks like some pretty solid work at first glance but please allow us a few days to provide a more thorough review. Thanks!

package.json
```diff
@@ -10,7 +10,7 @@
     "lint:fix": "eslint --fix --ignore-path .eslintignore --ignore-pattern webpack . && kacl lint",
     "lint": "eslint --ignore-path .eslintignore --ignore-pattern webpack . && kacl lint",
     "pretest": "rimraf ./.nyc_output",
-    "test": "ts-mocha -p tsconfig.json --recursive 'test/**/*.test.*'",
+    "test": "ts-mocha -p tsconfig.json --recursive 'test/**/*.test*'",
```
toddbaert, line 13: Fixes issue in recently merged PR that skips all tests with `tests`(plural) in there name.

toddbaert:
I made a few changes - namely, make `templates` an array, not an object, and required `.json` files for the directory definitions, to make things more similar to the way email templates work.

package.json
willvedd: Thanks for picking this out, however, it doesn't belong in this PR, I'll submit another PR to address specifically.

src/context/yaml/handlers/brandingTemplates.js
```diff
@@ -0,0 +1,57 @@
+import fs from 'fs-extra';
```
willvedd, line 1: Are you actually interested in creating a separate set of handlers for branding templates? It would seem to me that branding templates fall squarely within the realm of branding. And judging by the [deletion of the `templates` property](https://github.com/auth0/auth0-deploy-cli/pull/438/files#diff-1b948ca54d6b7880eb055d92d1c07525248d308e455ccae5660c01e4473065e3R33), it looks like there is a shared concern between these two handlers.

We typically try to conceptually tie the resource types in our config-management repos with the [Management API entities](https://auth0.com/docs/api/management/v2). So in this case, I'd really love to see it all bundled underneath a generalized branding umbrella if you can.

src/context/yaml/handlers/brandingTemplates.js
```diff
@@ -0,0 +1,57 @@
...
+  return {
+    branding: {
+      ...branding,
+      templates
+    }
+  };
+}
+
+async function dump(context) {
+  const { branding } = context.assets || { branding: undefined };
```
willvedd, line 29: Similarly to my point above, it looks like there's a shared concerns about the `branding` property between this handler and the existing branding handler. I think it would be sensical and more clear to combine together.

src/tools/auth0/handlers/brandingTemplates.js
```diff
@@ -0,0 +1,82 @@
...
+    super({
+      ...options,
+      type: 'branding'
+    });
+  }
+
+  async getType() {
+    // in case client version does not support branding
+    if (!this.client.branding || typeof this.client.branding.getUniversalLoginTemplate !== 'function') {
+      return {};
```
willvedd, line 32: Is it correct to return an empty object here? I'm mostly asking this out loud, not sure myself off the top of my head.

src/tools/auth0/handlers/brandingTemplates.js
willvedd: Is this requirement documented somewhere? I haven't personally tested new branding templates, so excuse my unfamiliarity.

willvedd (changes requested):
First off, great work and thanks for submitting this! It's been a requested feature in the past and serves as a good example PR for implementing Management API functionality.

It looks pretty sensible, I have a few smaller questions that I hope you can answer. My biggest piece of feedback is to combine this new feature with the existing branding handler. Conceptually we like to pair these resource types very closely with the Management API for consistency. Is this something you'd be able to accommodate?

package.json
willvedd: https://github.com/auth0/auth0-deploy-cli/pull/445/files

toddbaert: Cool, I'll rebase on this when it's merged.

src/context/yaml/handlers/brandingTemplates.js
toddbaert: 100%. Actually, I didn't realize that the templates were a "sub-entity" of branding until I'd already started, and then didn't ever re-evaluate the file structure.

I can easily just add the new template logic to the existing branding handler, and update the test filenames to be consistent as well. I don't anticipate any functional changes, so if it's OK with you I will just squash the change into the existing commit.

src/context/yaml/handlers/brandingTemplates.js
toddbaert: :100:

src/tools/auth0/handlers/brandingTemplates.js
toddbaert: Personally, I think throwing an exception might be a better choice, to make things less mysterious, however, this is the pattern used already in other handlers for missing client functionality, so I went the same route. I think if you want to change it, you might want to consider it a breaking change and do it in a separate PR.

src/tools/auth0/handlers/brandingTemplates.js
toddbaert: Yep, the API returns a 400 if there's no custom domain configured. There's some public doc here (see first paragraph): https://auth0.com/docs/customize/universal-login-pages/universal-login-page-templates

toddbaert:
Thanks for the review @willvedd . Please see my individual responses. I agree that this functionality can and should be combined with the branding handler. I will action that tomorrow, I dont think it'll be much effort.

src/tools/auth0/handlers/brandingTemplates.js
willvedd: Got it. Thanks!

src/context/yaml/handlers/brandingTemplates.js
willvedd: > Actually, I didn't realize that the templates were a "sub-entity" of branding until I'd already started

Agreed that the hierarchy isn't obvious, but it rids us of needing to make that decision ourselves.

> so if it's OK with you I will just squash the change into the existing commit

Yep, all good with me.

src/tools/auth0/handlers/brandingTemplates.js
willvedd: There are some hidden levers behind the scenes that control the type of tenant that's provisioned to a customer, so in certain cases, some customers tenants may not have certain features. I don't know if that's applicable for this feature specifically, but it's certainly possible. For that reason, fault-tolerance is preferred in this situation.

I'm mostly trying to understand the contract imposed by these handlers and if an empty object adequately describes a n empty configuration. But that's a discussion for another time. Since it's an existing pattern let's keep what you have here.

toddbaert:
@willvedd I think I've resolved your concerns. Let me know. I also see some other pending PRs, if you'd like to merge them first I can rebase on them.

ean:
Running dump with the latest revision of this PR fails on free accounts:
```
2022-03-18T07:55:04.655Z - info: Retrieving branding data from Auth0
2022-03-18T07:55:05.345Z - error: Problem running command export during stage load when processing type branding
2022-03-18T07:55:05.345Z - error: The account is not allowed to perform this operation, please contact our support team
```

I think there is a 403 that needs to be handled when calling `this.client.customDomains.getAll();`

willvedd:
>Let me know. I also see some other pending PRs, if you'd like to merge them first I can rebase on them.

I merged the testing fix earlier today. 

Re: the rest of this PR, I'm spending some time to perform some testing manually. The first thing I notice is that this feature creates an empty `branding_templates` directory in the case of no templates. Is this something you mind addressing?

toddbaert:
> > Let me know. I also see some other pending PRs, if you'd like to merge them first I can rebase on them.
> 
> I merged the testing fix earlier today.
> 
> Re: the rest of this PR, I'm spending some time to perform some testing manually. The first thing I notice is that this feature creates an empty `branding_templates` directory in the case of no templates. Is this something you mind addressing?

I will address this as well as @ean 's finding above about the 403 on free accounts.

toddbaert:
@ean @willvedd I pushed a fixup commit with the requested changes, and rebased on master.

src/tools/auth0/handlers/branding.js
```diff
@@ -11,28 +27,66 @@ export default class BrandingHandler extends DefaultHandler {
...
+      log.debug(`Error calling branding API, ${err.message}, status code: ${err.statusCode}`);
+      if (err.statusCode === 403) return branding;
+      if (err.statusCode === 404) return branding;
+      if (err.statusCode === 501) return branding;
       throw err;
     }
   }
 
   async processChanges(assets) {
     const { branding } = assets;
```
willvedd, line 65: I believe you need something like:

```
if (!branding) return;
```

Or else if the `branding` property is absent from assets, the `processChanges` method will choke further down the line.

willvedd:
@toddbaert Alright I was able to test locally for a number of situations and the only one I'm having issue with is if no branding configuration exists, and so the `processChanges` method throws an unhandled error. I mention a proposed fix above; with that conditional, I can confirm that all scenarios seem to work as intended. Once you're able to add that in, I think this is ready!

toddbaert:
Perfect. I'll do that in a couple hours!

toddbaert:
@willvedd I added the conditional you mentioned, as well as a new test for that case (https://github.com/auth0/auth0-deploy-cli/pull/438/files#diff-e09b4c99c3f517300a289e696063f1b8097225dfb392bbf183456aa75e2092dbR159) and I've squashed all the commits together.

willvedd (approved):
Great work and thank you for the contribution! Looking forward to having this in the next release.

PR #444, opened 2022-03-16
DXCDT-75: `calcChanges` function to Typescript

willvedd:
## ✏️ Changes

As part of the incremental Typescript migration, I've identified the `calcChanges` function as being something that would benefit from static typing. It is a dense, complex function that performs the fundamental diff'ing calculations for this project. 

I took my best stab at this, but it is not entirely complete by any means. One major open question (noted in the unit tests) is the "grouping" of identifiers. It appears that some test cases allow a mixed array of strings and string arrays, however, when typing with that union type, the code doesn't pass static type analysis. I've decided to _not_ refactor this code to prevent introducing breaking changes and instead will consider refactoring the `identifiers` type later in the future.

Another missing piece of note is more canonical `Asset` and `Handler` types. For now, I've just defined them within the scope of the `calculateChanges` function and will replace once the handlers get migrated to TS and lend themselves to a more canonical type.

One thing to point out is that I separated this function out into it's own file for organization and also renamed it to `calculateChanges` to differentiate it from the Auth0 API handlers that share a method with it's original name, `calcChanges`. This general reorganization makes it easier to navigate the codebase and unit test this specific function.

 

## 🎯 Testing

Tests remain the same, however, they're now written in Typescript.


src/tools/calculateChanges.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Temporary type for Auth0 API handler until more canonical types are created
+type Handler = {
+    id: string
+    name?: string
+    objectFields: string[]
+    objString: (arg0: Asset) => string
+
+}
```
willvedd, line 96: Temporarily defining these types within this file until more canonical types are generated from their respective TS migrations.

src/tools/calculateChanges.ts
```diff
@@ -0,0 +1,207 @@
...
+
+
+export function calculateChanges({ handler, assets, existing, identifiers = ['id', 'name'], allowDelete }: {
+    handler: Handler,
+    assets: Asset[],
```
willvedd, line 101: Changed the name and the function signature to be an object instead of a five-arrity function.

test/tools/calculateChanges.test.ts
```diff
@@ -0,0 +1,338 @@
...
+            { client_id: 'client1', audience: 'audience1', id: 'id3' },
+            { client_id: 'create1', audience: 'create1', id: 'create1' }
+        ];
+
+        const { create } = calculateChanges({
+            handler: mockHandler,
+            assets,
+            existing,
+            allowDelete: false,
+            //@ts-ignore need to investigate why this "grouping" exists
```
willvedd, line 55: This is one oddity that still needs to be understood, why we're passing in an array of mixed types. This test will pass if the nested array is broken down; not really sure what it's doing by looking at the code. 

Nevertheless, will not be refactoring, just something to point out.

adamjmcgrath (approved):

PR #446, opened 2022-03-17
DXCDT-74: JSON Handlers Typescript migration

willvedd:
## ✏️ Changes

Applying types to the JSON handlers, the primary abstraction for writing and parsing JSON configuration files for the respective resource types. The types are fairly straightforward, branching off a `DirectoryHandler<T>` generic type that inherits an expected type representing the parsed JSON file. 

For the most part there were no functional changes necessary to ensure that the compiler was satisfied, but I will note a few instances below.

## 🎯 Testing

This PR does not introduce any changes in tests.

### DIFF
--- a/src/context/directory/handlers/emailTemplates.ts
+++ b/src/context/directory/handlers/emailTemplates.ts
@@ -1,3 +1,4 @@
+// @ts-nocheck TODO: FIX TYPE ISSUES IN THIS FILE. DO NOT MERGE UNTIL FIXED!
 import fs from 'fs-extra';
 import path from 'path';
 import { constants, loadFileAndReplaceKeywords } from '../../../tools';
@@ -6,12 +7,17 @@ import log from '../../../logger';
 import {
   getFiles, existsMustBeDir, dumpJSON, loadJSON
 } from '../../../utils';
+import { DirectoryHandler } from '.'
 
-function parse(context) {
+type ParsedEmailTemplates = {
+  emailTemplates: unknown | undefined
+}
+
+function parse(context): ParsedEmailTemplates {
   const emailsFolder = path.join(context.filePath, constants.EMAIL_TEMPLATES_DIRECTORY);
   if (!existsMustBeDir(emailsFolder)) return { emailTemplates: undefined }; // Skip
 
-  const files = getFiles(emailsFolder, [ '.json', '.html' ]).filter((f) => path.basename(f) !== 'provider.json');
+  const files = getFiles(emailsFolder, ['.json', '.html']).filter((f) => path.basename(f) !== 'provider.json');
 
   const sorted = {};
 
@@ -41,8 +47,8 @@ function parse(context) {
   };
 }
 
-async function dump(context) {
-  const emailTemplates = [ ...context.assets.emailTemplates || [] ];
+async function dump(context): Promise<void> {
+  const emailTemplates = [...context.assets.emailTemplates || []];
 
   if (!emailTemplates) return; // Skip, nothing to dump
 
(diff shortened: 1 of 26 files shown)


src/context/directory/handlers/emailTemplates.ts
```diff
@@ -22,17 +27,24 @@ function parse(context) {
     if (ext === '.html') sorted[name].html = file;
   });
 
-  const emailTemplates = [];
-  Object.values(sorted).forEach((data) => {
-    if (!data.meta) {
-      log.warn(`Skipping email template file ${data.html} as missing the corresponding '.json' file`);
-    } else if (!data.html) {
-      log.warn(`Skipping email template file ${data.meta} as missing corresponding '.html' file`);
+  const emailTemplates = Object.values(sorted).flatMap(({
```
adamjmcgrath, line 30: Am assuming `flatMap` is ok because you don't support Node 10

src/context/directory/handlers/pages.ts
```diff
@@ -6,33 +6,41 @@ import log from '../../../logger';
...
-  });
+    if (!acc[name]) acc[name] = {};
+    if (ext === '.json') acc[name].meta = file;
+    if (ext === '.html') acc[name].html = file;
+    return acc
+  }, {});
 
-  const pages = [];
-  Object.values(sorted).forEach((data) => {
+  const pages = Object.values(sorted).map((data) => {
```
adamjmcgrath, line 34: `pages` may have undefined items in it now that you're using map

src/context/directory/handlers/emailTemplates.ts
willvedd: We haven't taken a hard stance on Node version support, but yes, based on [it's official support](https://endoflife.date/nodejs), it' safe to say that we're not supporting Node 10.

Although, now that you bring it up, it's certainly an important thing to consider. Everything transpiles into ES2016, but we'll have to double check how many previous versions of Node can adequately run that.

src/context/directory/handlers/pages.ts
willvedd: Good point, though I walked through and the `sorted` value is the result of a reduce, which at worse is an empty object. `Object.values({})` evaluates to an empty array, so I don't believe it will ever be undefined.

However, while doing so, I realized that the map function doesn't always return a value, so I've instead swapped it with `flatMap` to weed out any potentially undefined values it generates.

src/context/directory/index.ts
```diff
@@ -76,10 +93,7 @@ export default class {
 
     await Promise.all(Object.entries(handlers).map(async ([ name, handler ]) => {
       try {
-        const data = await handler.dump(this);
-        if (data) {
-          log.info(`Exporting ${name}`);
-        }
+        await handler.dump(this);
```
willvedd, line 96: `dump` returns void so `data` is not providing anything useful here.

src/context/directory/handlers/pages.ts
```diff
@@ -6,33 +6,41 @@ import log from '../../../logger';
...
-  });
+    if (!acc[name]) acc[name] = {};
+    if (ext === '.json') acc[name].meta = file;
+    if (ext === '.html') acc[name].html = file;
+    return acc
+  }, {});
 
-  const pages = [];
-  Object.values(sorted).forEach((data) => {
+  const pages = Object.values(sorted).flatMap((data) => {
```
adamjmcgrath, line 34: You still need to return empty arrays for the "if" and "else if" branches to prevent `undefined`s

adamjmcgrath (approved):

PR #452, opened 2022-03-22
DXCDT-73: YAML handlers in typescript

willvedd:
## ✏️ Changes

Similarly to #446 , this PR migrates all YAML handlers into Typescript. These handlers are the abstraction used to interface with YAML files, both for reading and writing. This PR only addresses the types and makes little attempt to refactor any inconsistent behavior (unless unavoidable). You will notice that there are several places where I leave `//TODO` comments of where code should be revisited and possibly refactored. However, for this first pass, this PR recognizes that not everything can be fixed right away and getting _some_ typing in place is still a win.

## 🎯 Testing

No changes to testing, but can confirm that tests do pass.


src/context/yaml/index.ts
```diff
@@ -61,14 +78,14 @@ export default class {
 
     // Run initial schema check to ensure valid YAML
     const auth0 = new Auth0(this.mgmtClient, this.assets, toConfigFn(this.config));
-    await auth0.validate('validate');
+    await auth0.validate();
```
adamjmcgrath, line 81: I assume this is intentional (removing the argument)

adamjmcgrath (approved):

src/context/yaml/index.ts
willvedd: Correct, the compiler was complaining because the `validate` method doesn't accept any arguments.

PR #455, opened 2022-03-24
DXCDT-86: Context Typescript migration

willvedd:
## ✏️ Changes

Context is a concept in the Deploy CLI for sharing data and configuration between YAML and JSON handlers without having a formal message bus to transmit. Although no formal `Context` type is proposed in this PR, it can be considered the union type between `DirectoryContext` and `YAMLContext` , the share context for their respective types of handlers. The concept of context spans wide across the codebase and appears in a bulk of function signatures, so typing context will result in tighter types types elsewhere in the codebase. 

As with the other TS migration PRs, I tried to resist the temptation to refactor as much as I could. You may notice some `//@ts-ignore`'s and TODO comments. The thinking is to get comprehensive typing coverage first, then refactor as necessary. 

## 🎯 Testing

No functional changes. All unit tests pass.


src/context/index.ts
```diff
@@ -4,9 +4,10 @@ import YAMLContext from './yaml';
 import DirectoryContext from './directory';
```
willvedd, line 4: This file is the bulk of the core changes for this PR.

src/context/index.ts
```diff
@@ -4,9 +4,10 @@ import YAMLContext from './yaml';
...
 
 import { isDirectory } from '../utils';
-import { version as packageVersion } from '../../package.json';
 import log from '../logger';
 
+const { version: packageVersion } = require('../../package.json')
```
willvedd, line 9: This was necessary to quell the TS compiler because the package.json file resides outside the purview of the tsconfig file.

src/context/index.ts
```diff
@@ -19,23 +20,67 @@ const nonPrimitiveProps = [
   'INCLUDED_PROPS'
 ];
 
-export default async function(config) {
+type Config = {
```
willvedd, line 23: First pass at typing the config object, just enough to get compilation to pass. There is another Jira ticket to flesh out configuration a bit more later.

src/context/index.ts
```diff
@@ -19,23 +20,67 @@ const nonPrimitiveProps = [
...
+  rules: Asset[],
+  rulesConfigs: Asset[],
+  tenant: Asset,
+  triggers: Asset[],
+  //non-resource types
+  exclude: {
+    [key: string]: string[]
+  },
+  clientsOrig: Asset[],
+}
```
willvedd, line 65: This `Assets` type is a bit unwieldy but it is fundamental to many of the core functions of this repo. Like the `Config` type, there is a task to clear this a bit more in the future.

src/context/index.ts
```diff
@@ -19,23 +20,67 @@ const nonPrimitiveProps = [
...
+
+export const setupContext = async (config: Config): Promise<DirectoryContext | YAMLContext> => {
```
willvedd, line 67: Named this function instead of relying on unnamed default export.

adamjmcgrath (approved):

PR #456, opened 2022-03-24
DXCDT-71: Auth0 API handler default class Typescript migration

willvedd:
## ✏️ Changes

The API handlers a key abstraction in this codebase responsible for applying configuration diffs by reading and writing data from Auth0's management API. Unlike some of the other Typescript migration PRs, this one only encompasses changes necessary to type the base class. I suspect that the actual API handlers that extend off this base class will require quite a few changes so best to separate.

I'll point out a few comments below but one point I'd like to mention is the addition of the `types.ts` file. I'm usually averse from arbitrarily separating code out by the file extension this way, but the more I go through this migration, the more I realize that there are several key types that are shared widely across the codebase and they don't necessarily fit into any one existing directory or file. IMO, this is usually a litmus test for unorganized code, but applying such a reorganization is out of the scope of this PR, thus, relying on a general `types.ts` file appears to be the best solution here.

## 🎯 Testing

No functional changes to the code, and tests pass. However, there was one test: `test/tools/auth0/handlers/rulesConfigs.tests.ts` that suddenly wasn't passing. I'm not sure the exact reason why this code prevents it from passing, but the change to get it passing was fairly straightforward and made me wonder if the test ever worked in the first place.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,0 +1,62 @@
+export type Auth0APIClient = {
+    [key: string]: {
+        getAll: any,
+        create: any,
+        update: any,
+        delete: any,
+    }
+    pool: any
+}// TODO: replace with a more accurate representation of the Auth0APIClient type 
+
+export type Config = {
+    AUTH0_DOMAIN: string
+    AUTH0_CLIENT_ID: string
+    AUTH0_CLIENT_SECRET: string
+    AUTH0_INPUT_FILE: string
+    AUTH0_ACCESS_TOKEN?: string
+    AUTH0_AUDIENCE?: string
+    AUTH0_API_MAX_RETRIES?: number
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -0,0 +1,62 @@
+export type Auth0APIClient = {
```
willvedd, line 1: I mentioned this in the PR description, but I'd like to call out again that I usually don't like to separate out types in this way. My preference is to keep the types as close to their functional code counterpart as possible. However, I discovered that these shared types are used widely across the codebase and didn't "belong" in any one location. And to avoid reorganizing the directories, I just opted for this general file. Over the course of this migration I'll be on the lookout for a better spot for these types, but I suspect this file is here for the long haul.

adamjmcgrath (approved):

PR #462, opened 2022-03-29
Adding `AssetType` for tighter enforcement of object keys

willvedd:
## ✏️ Changes

While implementing resource exclusions, the usages of objects with very specific keys correlating to asset types are the backbone of all the handlers. It became apparent that being able to enforce keys in an object with Typescript is crucial for the implementation. 

By leveraging this type, there will be less risk of accidentally not making crucial changes when adding or modifying new Management API features. In practice, this list would grow as features are added.

The type itself is a bit verbose but ensures a safe contract between various components in the system:

```ts
export type AssetTypes = 'rules' | 'rulesConfigs' | 'hooks' | 'pages' | 'databases' | 'clientGrants' | 'resourceServers' | 'clients' | 'connections' | 'tenant' | 'emailProvider' | 'emailTemplates' | 'guardianFactors' | 'guardianFactorProviders' | 'guardianFactorTemplates' | 'migrations' | 'guardianPhoneFactorMessageTypes' | 'guardianPhoneFactorSelectedProvider' | 'guardianPolicies' | 'roles' | 'actions' | 'organizations' | 'triggers' | 'attackProtection' | 'branding'
```

## 🎯 Testing

No functional changes so no additional tests.

### DIFF
--- a/src/context/directory/handlers/index.ts
+++ b/src/context/directory/handlers/index.ts
@@ -25,13 +25,14 @@ import attackProtection from './attackProtection';
 import branding from './branding';
 
 import DirectoryContext from '..'
+import { AssetTypes } from '../../../types'
 
 export type DirectoryHandler<T> = {
   dump: (context: DirectoryContext) => void,
   parse: (context: DirectoryContext) => T,
 }
 
-export default {
+const directoryHandlers: { [key in AssetTypes]: DirectoryHandler<{ [key: string]: unknown }> } = {
   rules,
   rulesConfigs,
   hooks,
@@ -57,4 +58,6 @@ export default {
   triggers,
   attackProtection,
   branding
-};
+}
+
+export default directoryHandlers;
--- a/src/context/yaml/handlers/actions.ts
+++ b/src/context/yaml/handlers/actions.ts
@@ -22,7 +22,8 @@ function parseCode(context: YAMLContext, code: string) {
 
 async function parse(context: YAMLContext): Promise<ParsedActions> {
   // Load the script file for each action
-  if (!context.assets.actions) return { actions: [] };
+  //@ts-ignore TODO: understand if empty array is intentionally being returned
+  if (!context.assets.actions) return [];
   const actions = {
     actions: [
       ...context.assets.actions.map((action) => ({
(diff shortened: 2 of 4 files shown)


src/context/yaml/handlers/actions.ts
```diff
@@ -22,6 +22,7 @@ function parseCode(context: YAMLContext, code: string) {
 
 async function parse(context: YAMLContext): Promise<ParsedActions> {
   // Load the script file for each action
+  //@ts-ignore TODO: understand if empty array is intentionally being returned
```
willvedd, line 25: This is an example of a compilation error that may be the source of a bug. In this case, we'd expect the return type to be an object with the key `actions`.

src/types.ts
```diff
@@ -59,4 +59,6 @@ export type Assets = {
         [key: string]: string[]
     },
     clientsOrig: Asset[],
-}
\ No newline at end of file
+}
+
+export type AssetTypes = 'rules' | 'rulesConfigs' | 'hooks' | 'pages' | 'databases' | 'clientGrants' | 'resourceServers' | 'clients' | 'connections' | 'tenant' | 'emailProvider' | 'emailTemplates' | 'guardianFactors' | 'guardianFactorProviders' | 'guardianFactorTemplates' | 'migrations' | 'guardianPhoneFactorMessageTypes' | 'guardianPhoneFactorSelectedProvider' | 'guardianPolicies' | 'roles' | 'actions' | 'organizations' | 'triggers' | 'attackProtection' | 'branding'
```
willvedd, line 64: The primary addition of this PR.

src/context/directory/handlers/index.ts
```diff
@@ -25,13 +25,14 @@ import attackProtection from './attackProtection';
...
 import DirectoryContext from '..'
+import { AssetTypes } from '../../../types'
 
 export type DirectoryHandler<T> = {
   dump: (context: DirectoryContext) => void,
   parse: (context: DirectoryContext) => T,
 }
 
-export default {
+const directoryHandlers: { [key in AssetTypes]: DirectoryHandler<{ [key: string]: unknown }> } = {
```
willvedd, line 35: The addition of `key in AssetTypes` ensures that only and all keys in the object match to strings in the `AssetTypes` union type.

src/types.ts
willvedd: @sergiughf please excuse the lack of a new line EOF, I'm making a linter change #461 🙈

PR #467, opened 2022-03-31
Typing Auth0 class [2/3]

willvedd:
## ✏️ Changes

Adding types to the nondescript `Auth0` class. This class is an amalgamation of the API handlers (still yet to be migrated), the Management API client and configuration. This is the class that ultimately gets used to invoke the fundamental process of loading all assets from Auth0 and apply proposed changes.


src/tools/auth0/handlers/index.ts
```diff
@@ -1,3 +1,4 @@
+//TODO:
```
willvedd, line 1: Complete TODO comment. Should read `//TODO: don't use wildcard imports`

src/tools/auth0/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  assets: Assets
+  handlers: APIHandler[]
+
+  constructor(client: Auth0APIClient, assets: Assets, config: ConfigFunction) {
+    this.client = pagedClient(client);
+    this.config = config;
+    this.assets = assets;
+
+    this.handlers = Object.values(handlers).map((h) => {
+      //@ts-ignore because prompts don't appear to have been universally implemented yet
```
willvedd, line 43: TODO: fix comment, shouldn't' be about prompts.

sergiughf (approved):

Issue #451, opened 2022-03-22
Proposal: Resource Exclusion


willvedd:
- **Start Date**: March 14, 2022
- **Github Issue**: [Proposal: Resource Exclusion · Issue #451](https://github.com/auth0/auth0-deploy-cli/issues/451)

# Summary
We propose a mechanism for excluding entire resources from the management purview of Deploy CLI.

Resource exclusion is an often requested feature from the community and has many legitimate use-cases. Currently, there is some resource-specific exclusion functionality, on a per-ID basis, but nothing that excludes entire resource types.

# Why Exclude?
Excluding entire resources are useful for developers who work on tenants with many of a certain type. In the case of hundreds or thousands of a type of resource, it may be operationally inconvenient to manage all of those configurations.

Additionally, exclusion is useful for organizations who provision certain resource types manually or through different API means, and they do not wish to automate that resource type under the purview of deploy CLI.

It is understood that exclusions operate bi-directionally, meaning they work both on import and export. Currently, there is no clear use case for excluding resources for only one direction.

# Existing Methods of Exclusion <a href="#user-content-deprecated-exclusion-props" id="deprecated-exclusion-props">#</a>
The following are existing configuration values for excluding resources on a per-ID basis:

- `AUTH0_EXCLUDED_RULES`
- `AUTH0_EXCLUDED_CLIENTS`
- `AUTH0_EXCLUDED_DATABASES`
- `AUTH0_EXCLUDED_CONNECTIONS`
- `AUTH0_EXCLUDED_RESOURCE_SERVERS`

There are **key issues** these configurations though:

1. They only apply on a per-ID basis, meaning that there is no way to automatically exclude all of that certain resource. This is a minor inconvenience for a handful of a certain resource, however, in the case of hundreds or thousands of a certain resource, it becomes untenable. Further, as more of a certain resource gets added, they'll necessarily need to be added to the Deploy CLI configuration, requiring more touch points.

2. This only covers a subset of all resources. The Auth0 Management API is still expanding in scope, and exclusions for each resource  Granted, it is possible to accommodate all n number of resource types with n number of exclusion properties, but can become unmanageable

3. Many entities may rely upon another specific resource in a complex web of dependencies. Each specific dependency is difficult to manage within the tool. For the simplicity and reliability of operation, it would be better to remove per-ID exclusions in favor of broader exclusions that can be more predictably managed.

# Deprecation

[...]

Issue #427, opened 2022-03-03
Allow skipping export of Organizations


nathanKramer:
### Describe the problem you'd like to have solved

Hi, I've recently updated the CLI to version 7, to allow exporting actions.

However, I'm having an issue where our tenant contains thousands of orgs.

The CLI now wants to export them along with their connections, which can't be done in a timely manner, and is not desired for our use-case.

My preference is to treat these Orgs as data, rather than as config that should be exported.
I'm wondering if there is any support for this. 
(or if there are any suggestions for ergonomic work-arounds)

I can't find any evidence that this is supported, but I thought I would ask.

It's not clear to me whether it is safe to try to delete organizations from the resulting export, and then deploy, or if this would delete organizations. (This issue seems relevant: https://github.com/auth0/auth0-deploy-cli/issues/387)

### Describe the ideal solution

Some way to exclude organizations from the export.

e.g, an environment variable, or some kind of manifest that outlines the desired export)

### Attempted work arounds

I tried dropping the `read:organizations` scope, but the export throws:

```
2022-03-03T03:21:01.811Z - info: Retrieving organizations data from Auth0
2022-03-03T03:21:02.012Z - error: Problem running command export during stage load when processing type organizations
2022-03-03T03:21:02.012Z - error: Insufficient scope, expected any of: read:organizations
2022-03-03T03:21:02.013Z - debug: Forbidden: Insufficient scope, expected any of: read:organizations
```

willvedd:
Thanks for raising this. I think this is a duplicate of #387 which I have just [merged a relevant fix](https://github.com/auth0/auth0-deploy-cli/pull/388) for. In short, excluding resources from the configuration-as-code files should exclude them from being handled by the deploy CLI. That is, as long as the data doesn't explicitly define those resources as an empty set and rather not defined at all, they should be skipped. I think what you've found is that organizations paired with the YAML format doesn't follow this pattern and that will be fixed in an upcoming release (#434).

But something to remember is that as long as `AUTH0_ALLOW_DELETE` isn't set to true, you're never at risk of deleting any of your organizations. So as far as workarounds go, until 7.5.0 is released, it's ok to remove organization configuration and have `AUTH0_ALLOW_DELETE:false` in your config, you'll just see some warnings.

willvedd:
[...]

Issue #83, opened 2019-01-31
Excluding clients in production


drjaydenm:
Hi there,
Great work on this CLI tool, it makes managing configurations across tenants really easy.

We are using the directory format with JSON files and have created the JSON config required for our API Test Clients. This is all working correctly. However these API Test Clients get deployed out to production like everything else in the repo.

I can see there are options for excluding certain rules and resource servers, however nothing exists for clients. We were wondering how you would advise managing API Test Clients across environments using the CLI tool.

Is it as simple as adding an exclude option to this CLI tool for excluding clients too? Or is there something simple we are missing here?

Thanks in advance for your help

johankvint:
We would really like this feature too.
In our test tenants we have additional clients setup to be used for generating tokens in different ways that we do not want in production.

fyockm:
published as `auth0-deploy-cli@3.1.0`

PR #468, opened 2022-03-31
DXCDT-84: Resource Exclusion

willvedd:
## ✏️ Changes

Introducing the initial implementation for resource exclusion. As proposed in #451 , this provides a mechanism for users to ignore entire resource types (organizations, connections, hooks, etc.). This mechanism functions bi-directionally and will completely omit all resources from being fetched, written to config-as-code and updating.

In practice, users will add a `AUTH0_EXCLUDED` property to their user-defined configuration file and list the resource types they wish to exclude from the purview of the tool.

Example:
```json
{
  "AUTH0_DOMAIN": "test.us.auth0.com",
  "AUTH0_CLIENT_ID": "FOO",
  "AUTH0_CLIENT_SECRET": "foobarsecret",
  "AUTH0_EXCLUDED": [
    "pages",
    "organizations",
    "clients"
   ]
}
```

The approach here is fairly naive but effective. Not all permutations of exclusions have been tested and because of inter-dependencies between resources, exclusions _may_ lead to unexpected behavior. But it's expected to function under the vast majority of use cases.

## 🔗 References

- [Resource exclusion RFC](#451)
- #83 
- #388 
- #427 

## 🎯 Testing

I fought with the existing unit tests and was only able to add a unit test to the Auth0 class side. This test asserts that no excluded resources get fetched or pushed to Auth0. However, I had difficulty implementing 

Regardless, this being a fairly wide-reaching concept, it probably isn't appropriate to unit test. Instead, we should have a broader set of integration that can better assert the behavior of this functionality.

Regardless, plenty of manual test were done on this to confirm that this works for both YAML and directory config types, and varying permutations of this configuration.


src/context/directory/index.ts
```diff
@@ -78,14 +78,17 @@ export default class DirectoryContext {
     // Copy clients to be used by handlers which require converting client_id to the name
     // Must copy as the client_id will be stripped if AUTH0_EXPORT_IDENTIFIERS is false
     //@ts-ignore because assets haven't been typed yet TODO: type assets
-    this.assets.clientsOrig = [...this.assets.clients];
+    this.assets.clientsOrig = [...this.assets.clients || []];
```
willvedd, line 81: Clients could theoretically be excluded and thus be empty so need to default to empty array.

src/context/directory/index.ts
```diff
@@ -78,14 +78,17 @@ export default class DirectoryContext {
...
 
     // Optionally Strip identifiers
     if (!this.config.AUTH0_EXPORT_IDENTIFIERS) {
       this.assets = stripIdentifiers(auth0, this.assets);
     }
 
-    await Promise.all(Object.entries(handlers).map(async ([name, handler]) => {
+    await Promise.all(Object.entries(handlers).filter(([handlerName]: [AssetTypes, DirectoryHandler<any>]) => {
```
willvedd, line 88: This array filter function is the primary mechanism for exclusion.

sergiughf (approved):

Issue #465, opened 2022-03-31
v7.6.0: Payload validation error: 'Expected type integer but found type null' on property idle_session_lifetime_in_minutes


larrydahooster:
### Description

_Bad Request: Payload validation error: 'Expected type integer but found type null' on property idle_session_lifetime_in_minutes (Force a user to login after they have been inactive for the specified number (unit: minutes)). (also) Payload validation error: 'Expected type integer but found type null' on property
 session_lifetime_in_minutes (Login session lifetime, how long the session will stay valid (unit: minutes))._

Somehow the dump and deploy functions are incompatible. In my case the dump does not add a "idle_session_lifetime_in_minutes" to my local dumped files. **I assume that the values 'session_lifetime' and 'idle_session_lifetime' do not get exported into the dump when they still have the untouched default values 4320 and 10080. Once I appended one second, the values showed up in the dump.**

### Reproduction

In my case it is a simple dump of my dev environment and a deploy to my staging environment. It is mandatory that the 'session_lifetime' and 'idle_session_lifetime' are untouched, in order to not be exported.

### Environment

> Please provide the following:

- **Version of this library used: 7.6.0**

pablotp:
Thanks @larrydahooster for reporting this.
I'm also affected.
- Auth0 configuration as JSON files exported using the a0deploy version 7.5.2
- When importing it with 7.6.0 (without applying any change) it returns the same error message.
For now, I've downgraded to 7.5.2 as workaround.

Floppy:
I've got around this by setting the following in my `tenant.json`. The `*_in_minutes` versions suggested by the error message didn't work, looks like it needs to be in hours. 🤷🏻 

```
  "idle_session_lifetime": 72,
  "session_lifetime": 180
```

willvedd:
Thanks for opening this. I think the root cause here is folks relying on the default session durations, but somewhere in the Deploy CLI's process, it converts those undefined values to null values that the Management API balks at. Can you folks **confirm that you are/were relying on these default values**?

Both `session_lifetime_in_minutes` and `idle_session_lifetime_in_minutes` are undocumented but supported by the Auth0 Management API when updating those values. On the GET, the values are in hours. There are some legacy use cases where decimal values were used to configure these settings, so the Deploy CLI needs to convert into minutes.

[...]

PR #95, opened 2019-02-26
convert non-integer `session_lifetime` to minutes

fyockm:
Auth0 Deploy CLI exports and imports are throwing an error when they configure a non whole hour value for `idle_session_lifetime` and `session_lifetime` in the Dashboard.

Err message looks like:

```
2019-02-18T19:13:37.297Z - error: Problem running command import during stage processChanges when processing type tenant
2019-02-18T19:13:37.297Z - error: Payload validation error: 'Expected type integer but found type number' on property idle_session_lifetime (Force a user to login after they have been inactive for the specified number (unit: hours)).
```
and
```
Payload validation error: 'Expected type integer but found type number' on property session_lifetime 
(Login session lifetime, how long the session will stay valid (unit: hours)).
```

## ✏️ Changes

Allow deploy-cli users to account for non-integer values for `session_lifetime` and `idle_session_lifetime` in the form of `session_lifetime_in_minutes` and `idle_session_lifetime_in_minutes`.
  
## 🔗 References
  
Slack:
- https://auth0.slack.com/archives/C9170558S/p1550518374059300
- https://auth0.slack.com/archives/CELNVJVH7/p1550235309035600
  
## 🎯 Testing
  
1. Set up Auth0 Deploy CLI
2. Configure tenant’s  “session_lifetime: 61” and “idle_session_lifetime: 61" (input a value NOT in whole hours(60/120/180))
3. a0deploy export -c config.json -f yaml -o ~/auth0cli.yaml
4. a0deploy import -c config.json -i tenant.yaml

✅ This change has unit test coverage
  
✅ This change has integration test coverage
  
🚫 This change has been tested for performance
  
## 🚀 Deployment
  
✅ This can be deployed any time

PR #471, opened 2022-04-04
Fix: Session lifetime values not converted if not explicitly set

willvedd:
## ✏️ Changes

This PR addresses the following error:

```shell
2022-04-04T18:01:35.392Z - error: Problem running command import during stage processChanges when processing type tenant
2022-04-04T18:01:35.392Z - error: Payload validation error: 'Expected type integer but found type null' on property session_lifetime_in_minutes (Login session lifetime, how long the session will stay valid (unit: minutes)).
```

This error occurs when attempting to import exported tenant configuration where `session_lifetime` and `idle_session_lifetime` values have not been set explicitly defined and instead rely on the default values. These values are converted into `session_lifetime_in_minutes` and `idle_session_lifetime_in_minutes` properties respectively in order to facilitate backwards compatibility and some legacy use cases. The bug is occurring during this conversion, where an undefined value is being converted into a null value. This null value then fails the Auth0 Management API validation.

I was only able to recreate when exporting to the directory format (JSON) and upon a fresh tenant; I'm unfamiliar with a way to unset those session duration values to inherit the defaults.

This PR fixes this issue by being a bit more deliberate with the calculations and only performing them if the original session duration value(s) are set initially, otherwise they are ignored.

## 🔗 References

Original Github issues: #465 
Original implementation: #95 

## 🎯 Testing

This new method of session duration conversion happens in a `sessionDurationsToMinutes` is unit tested to assure that absence of one or both properties doesn't result in an inappropriate conversion.

Another set of cases have been added to the YAML and Directory tenant handlers to assert that missing session duration configurations don't result in an error.

Lastly, quite a bit of manual testing was done with this on a fresh tenant. After recreating the issue can confirm that this fixes the issue.


src/utils.js
```diff
@@ -103,11 +103,6 @@ export function sanitize(str) {
   return sanitizeName(str, { replacement: '-' });
 }
 
-export function hoursAsInteger(property, hours) {
-  if (Number.isInteger(hours)) return { [property]: hours };
-  return { [`${property}_in_minutes`]: Math.round(hours * 60) };
```
willvedd, line 108: This function alway read a bit weird to me and I think that the root of this bug might reside in line 108 but unable to confirm 100%.

src/sessionDurationsToMinutes.ts
```diff
@@ -0,0 +1,16 @@
+function hoursToMinutes(hours: number): number {
+    return Math.round(hours * 60)
+}
+
+export const sessionDurationsToMinutes = (
```
willvedd, line 5: This is the new function that converts `session_lifetime` and `idle_session_lifetime` into their minute counterparts. 

It is most easily consumed when destructuring the result as can be seen in the handlers.

src/sessionDurationsToMinutes.ts
```diff
@@ -0,0 +1,16 @@
...
+    { session_lifetime, idle_session_lifetime }: { session_lifetime?: number, idle_session_lifetime?: number }): { session_lifetime_in_minutes?: number, idle_session_lifetime_in_minutes?: number } => {
+    const sessionDurations: {
+        session_lifetime_in_minutes?: number
+        idle_session_lifetime_in_minutes?: number
+    } = {};
+
+    if (!!session_lifetime) sessionDurations.session_lifetime_in_minutes = hoursToMinutes(session_lifetime)
```
willvedd, line 12: This conditional is what prevents this null bug from occurring. If either of the session durations are undefined, they won't be returned.

sergiughf (approved):

PR #472, opened 2022-04-04
Typing primary index.ts file

willvedd:
## ✏️ Changes

Applying types to the root `index.js` file (now `.ts`). The bulk of this work was typing the parameters both through inference and by explicit definition and applying them. I also moved the `ExportParams` and `ImportParams` closer to their core definitions and have them imported elsewhere.

## 🎯 Testing

No functional changes. Passed my manual tests.


src/args.ts
```diff
@@ -84,6 +109,7 @@ function getParams() {
     .example('$0 deploy -c config.json -i path/to/files', 'Deploy Auth0 via Path')
     .epilogue('See README (https://github.com/auth0/auth0-deploy-cli) for more in-depth information on configuration and setup.')
     .wrap(null);
+  //@ts-ignore because we know these types are either ExportParams or ImportParams. TODO: fix more native way of inferring types from yargs
```
willvedd, line 112: It's a real shame that `yargs` doesn't provide a simpler, more native way of inferring types. I can assure with certainty that the types I've set in the function signature are correct and the ones auto-generated are wrong.

src/index.ts
```diff
@@ -23,9 +27,14 @@ async function run(params) {
     bootstrap();
   }
 
-  log.debug(`Start command ${params._[0]}`);
-  await cmd(params);
-  log.debug(`Finished command ${params._[0]}`);
+  log.debug(`Start command ${command}`);
+  if (['deploy', 'import'].includes(command) && 'input_file' in params) {
```
willvedd, line 31: Breaking out the one-liner `await cmd(params)` into a few more lines making it explicitly clear which commands are being executed.

Also, the `'input_file' in params` and `'output_folder' in params` conditionals are necessary to convince the compiler that the correct parameter type is getting passed through.

src/index.ts
```diff
@@ -35,28 +44,33 @@ if (require.main === module) {
 
   log.debug('Starting Auth0 Deploy CLI Tool');
 
-  // Set log level
-  log.transports.console.level = params.level;
```
willvedd, line 39: I did make this small deletion, but it's fairly clear that `params.level` isn't getting set anywhere. I think it is a relic of an older refactor. Not worth the noise to keep.

sergiughf (approved):

PR #480, opened 2022-04-06
Applying types to `src/tools/auth0/client.js` file

willvedd:
## ✏️ Changes

Migrating the `src/tools/auth0/client.js` file to Typescript. This file primarily serves to apply pagination support to the Auth0 Node library and orchestrate the large number of requests that this library generates.


src/tools/auth0/client.ts
```diff
@@ -61,10 +73,14 @@ function checkpointPaginator(client, target, name) {
   };
 }
 
-function pagePaginator(client, target, name) {
-  return async function (...args) {
+function pagePaginator(
+  client: Auth0APIClient,
+  target,
```
willvedd, line 78: There's still a bit of mystery what this `target` argument is or should be doing. From what I can tell, it's the Auth0 entity manager in the Node Auth0 library, but it's highly complicated to type ourselves as consumers.

sergiughf (approved):

Issue #286, opened 2020-11-02
Document all possible config keys


avnerenv0:
### Describe the problem you'd like to have solved

I stumbled on the `AUTH0_EXCLUDED_CONNECTIONS` configuration property and it really helped me use this library better. But I really found it by accident. I'd like all config options to be documented.

### Describe the ideal solution

Document all config options in the README (or a linked MD file)

okeegan-myob:
You have my +1 (though if I could +100 to this, I would). I really want to understand what `AUTH0_ALLOW_DELETE` will actually do. I understand it to be like the aws s3 sync with --delete but it would be great to have it explicitly described.

bfricka:
At the bare minimum, create a public JSON schema. This is good for the project as well since it can be used for validation. Considering the absolute neglect of the issues in the project, it's hard to see it happening anytime soon, though.

willvedd:
Update: this is being worked on right now (better late than never)! Looking forward to sharing with folks very soon.

willvedd:
This has been addressed with the addition of documentation to this project, including a [Configuring the Deploy CLI page](https://github.com/auth0/auth0-deploy-cli/blob/master/docs/configuring-the-deploy-cli.md). Please check it out and create PRs with any suggested alterations. Thanks for everyone's patience!

PR #494, opened 2022-04-14
Adding obfuscation of sensitive values through `sensitiveFieldsToObfuscate`

willvedd:
## ✏️ Changes

Adding a class `sensitiveFieldsToObfuscate` constructor property that defines a set of potentially sensitive values like keys and tokens that should be obfuscated when creating the configuration files. On the reverse side, these sensitive values will be removed from the payload when creating and updating so that they're not overriden with the placeholder value.

This functionality is set to be used with the upcoming log streams support and many more in the future. It also has the potential, from what I can tell, to replace the "excluded defaults" abstraction which has confused folks in the past (#286 ).

## 🎯 Testing

Added unit tests as well as higher-up integration tests.


sergiughf (approved):
Thanks for always making sure every functionality we add is thoroughly covered with tests <3 They also help me understand further the implementation.

PR #447, opened 2022-03-18
Import/export support for branding with directory as source/target.

ean:
## ✏️ Changes

Adds support for importing and exporting branding data when using directory as target/source.

This change depends on https://github.com/auth0/auth0-deploy-cli/pull/438, I will rebase and update when it has been merged. But I wanted to get some initial feedback on this.

## 🎯 Testing

> Describe how this can be tested by reviewers. Please be specific about anything not tested and reasons why.
branding/branding.json will be written to output directory when using `directory` output format.

✅ This change has unit test coverage

PR #505, opened 2022-04-19
DXCDT-48: Directory branding support

willvedd:
## ✏️ Changes

Inspired by #447 , this PR adds support for managing branding settings with the directory configuration format (YAML already supported).

Example of resulting **branding.json**:
```json
{
  "colors": {
    "primary": "#F8F8F2",
    "page_background": "#222221"
  },
  "font": {
    "url": "https://mycompany.org/font/myfont.ttf"
  },
  "logo_url": "https://foo.com/assets/logo.png"
}

```

A good bit of this PR is massaging the tests to reflect this new support and also enabling a high level of fault tolerance with regards to dumping and processing the data in the case of undefined properties.

## 🔗 References

Original inspiration PR: #447

### DIFF
--- a/src/context/directory/handlers/branding.ts
+++ b/src/context/directory/handlers/branding.ts
@@ -51,12 +51,13 @@ async function dump(context: DirectoryContext) {
   if (!!templates) dumpBrandingTemplates(context);
 }
 
-const dumpBrandingTemplates = ({
-  filePath,
-  assets: {
+const dumpBrandingTemplates = ({ filePath, assets }: DirectoryContext): void => {
+  if (!assets || !assets.branding) return;
+
+  const {
     branding: { templates = [] },
-  },
-}: DirectoryContext): void => {
+  } = assets;
+
   const brandingTemplatesFolder = path.join(
     filePath,
     constants.BRANDING_DIRECTORY,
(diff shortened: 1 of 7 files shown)


sergiughf (approved):

willvedd:
@ean this has been included in the 7.10.0. Thank you for your help and patience!

src/context/directory/handlers/branding.ts
```diff
@@ -10,9 +10,14 @@ type ParsedBranding = {
 };
 
 function parse(context: DirectoryContext): ParsedBranding {
+  const brandingDirectory = path.join(context.filePath, constants.BRANDING_DIRECTORY);
+
+  if (!existsMustBeDir(brandingDirectory)) return { branding: undefined };
+
+  const branding = loadJSON(path.join(brandingDirectory, 'branding.json'), context.mappings);
```
tpires, line 17: this will break a few integrations if the tenant doesn't have No-Code Editor feature enabled. `branding` directory might exist for Page Template feature, but the `branding.json` doesn't. If the file doesn't exists, this should return `{ branding: undefined }`.

tpires: this is what happens on accounts that don't have this feature enable, but use universal login page template.
image

willvedd: @tpires Please open a separate issue outlining the problem. Thanks.

PR #512, opened 2022-04-26
Standardize empty vs null resources (1 of 2)

willvedd:
## ✏️ Changes

This PR standardizes the treatment of empty and null resources. This subtle distinction is a very important one to make. Consider: 

**Empty:** The actual and intentional empty set of a particular resource. A fresh tenant will have several empty resource types when initialized. Empty sets may also be used to delete all of a particular resource on a tenant. Can best be thought of as an empty array, example: `{ hooks: [] }`

**Null:** The absence, omission or otherwise unintentional emptiness of a set. The omission of a resource may indicate the intentional omission of a resource from the purview of management. It could also indicate some 
Can best be thought of as a null value, example: `{ hooks: null }`

The current codebase heavily relies on the practice of explicitly defining values as `undefined` to signal an empty value, but this practice is not fully consistent. Some other resource types will rely on `void`, `null`, `{}` and `[]` to signify null values. The large deviation in approach of distinguishing between null and empty resource sets makes it difficult to apply singular code treatments to all resource types at once.

## 🎯 Testing

Changes in testing to assert that `null` values are returned instead of `undefined`.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -176,41 +176,43 @@ export type Config = {
 export type Asset = { [key: string]: any };
 
 export type Assets = {
-  actions: Asset[];
-  attackProtection: Asset;
-  branding: Asset;
-  clients: Asset[];
-  clientGrants: Asset[];
-  connections: Asset[];
-  databases: Asset[];
-  emailProvider: Asset;
-  emailTemplates: Asset[];
-  guardianFactorProviders: Asset[];
-  guardianFactors: Asset[];
-  guardianFactorTemplates: Asset[];
+  actions: Asset[] | null;
+  attackProtection: Asset | null;
+  branding: {
+    templates?: { template: string; body: string }[] | null;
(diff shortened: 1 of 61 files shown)


src/types.ts
```diff
@@ -175,43 +175,45 @@ export type Config = {
...
-  clients: Asset[];
-  clientGrants: Asset[];
-  connections: Asset[];
-  databases: Asset[];
-  emailProvider: Asset;
-  emailTemplates: Asset[];
-  guardianFactorProviders: Asset[];
-  guardianFactors: Asset[];
-  guardianFactorTemplates: Asset[];
+export type Assets = Partial<{
```
willvedd, line 178: Because `context.assets` can have undefined properties due to omission in the config files, the types should reflect that possibility with the `Partial` utility type.

sergiughf (approved):

PR #524, opened 2022-05-03
Fix: Adding IDs to newly-created actions for successful deploy

willvedd:
## ✏️ Changes

Actions that are created through the Deploy CLI always fail to deploy, this happens because they're created from configurations that lack an ID, but an ID is required to subsequently deploy a newly-created action. What this results in is the `deployAction` function having an `undefined` ID and always failing. This PR introduces a step that pairs newly-created actions with their ID so that they'll send the correct, newly-minted ID along to the `deployAction` function.

Also, during testing I observed a few instances where an action took ~30-45 seconds to reach the "built" state, a prerequisite for deployment, so I upped the effective timeout to 2 minutes.

## 🔗 References

[ESD-19344](https://auth0team.atlassian.net/browse/ESD-19344) - Customer intermittently observing "Bad Request: A draft must be in the built' state before it can be deployed." error when using auth0-deploy-cli

## 🎯 Testing

Added unit test case to ensure that deploy mechanism receives an ID. Also verified through manual testing, going from clearing out all actions to deploying several, back and forth.

### DIFF
--- a/src/tools/auth0/handlers/actions.ts
+++ b/src/tools/auth0/handlers/actions.ts
@@ -3,6 +3,7 @@ import DefaultAPIHandler, { order } from './default';
 import log from '../../../logger';
 import { areArraysEquals } from '../../utils';
 import { Asset } from '../../../types';
+import { showCompletionScript } from 'yargs';
 
 const MAX_ACTION_DEPLOY_RETRY_ATTEMPTS = 60; // 60 * 2s => 2 min timeout
 
@@ -215,7 +216,7 @@ export default class ActionHandler extends DefaultAPIHandler {
       ...changes.create
         .filter((action) => action.deployed)
         .map((actionWithoutId) => {
-          // Supplement the just-created actions with their IDs
+          // Add IDs to just-created actions
           const actionId = postProcessedActions?.find((postProcessedAction) => {
             return postProcessedAction.name === actionWithoutId.name;
           })?.id;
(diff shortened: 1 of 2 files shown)


src/tools/auth0/handlers/actions.ts
```diff
@@ -131,10 +132,10 @@ export default class ActionHandler extends DefaultAPIHandler {
           log.info(`[${this.type}]: Waiting for build to complete ${this.objString(action)}`);
           action.retry_count = 1;
         }
-        if (action.retry_count > MAX_ACTION_DEPLOY_RETRY) {
+        if (action.retry_count > MAX_ACTION_DEPLOY_RETRY_ATTEMPTS) {
           throw err;
         }
-        await sleep(1000);
+        await sleep(2000);
```
willvedd, line 138: From my limited testing, I found that an action either built very quickly, almost immediately or took a long time, I believe retrying every second is unnecessary.

src/tools/auth0/handlers/actions.ts
```diff
@@ -203,11 +204,33 @@ export default class ActionHandler extends DefaultAPIHandler {
...
+            return postProcessedAction.name === actionWithoutId.name;
+          })?.id;
+
+          const actionWithId = {
+            ...actionWithoutId,
+            id: actionId,
+          };
+          return actionWithId;
+        })
+        .filter((action) => !!action.id),
```
willvedd, line 230: This is the bulk of this PR. Immediately after processing changes through `super.processChanges(...)`, re-fetch the actions to receive the newly-created IDs and match those up with with the existing actions objects for deployment.

sergiughf (approved):

PR #530, opened 2022-05-05
DXCDT-92: Adding prompts support

willvedd:
## ✏️ Changes

As requested in #229 , this PR adds support for prompts implementing both the [prompts settings](https://auth0.com/docs/api/management/v2#!/Prompts/get_prompts) and the [custom text for a prompt](https://auth0.com/docs/api/management/v2#!/Prompts/get_custom_text_by_language). Previously, there was support for prompts custom settings for the YAML format, but nothing for directory. The custom text capability is a more important management burden to add to the tool.

When dumping, the tool will fetch the tenant settings and see which languages are enabled and then use those to requests all custom text for those languages.

**Note about performance:** This addition incurs a fairly high performance penalty when dumping. I discuss this a bit in a [comment soliciting prompts feedback](https://github.com/auth0/auth0-deploy-cli/issues/229#issuecomment-1116700163), but the reason is that the custom text endpoint requires a separate request per-screen per-language. There are ~25 screens so if a tenant is supporting many languages, this could easily turn into hundreds of additional requests. However, most folks acknowledged that performance was worth the management capability itself. 

**Note about reconciliation:**  Because the pattern of the custom text endpoint differs so much from most other Auth0 management API endpoints, it does not immediately integrate with the diff'ing functionality already built into the tool. The task to implement that is quite high and has been deferred to a future point in time. What this means is that the tool only supports creates and updates, but **not deletions of custom text** . I don't believe this to be a major use case because the presence of those configurations when not used, should not impact any other functionality. If this deletion capability is ever requested, the level of effort might be high, but it can nonetheless be added at a later point in time.

## 🔗 References

Original Github issue: #229 

## 🎯 Testing

Added unit tests for the Auth0 Prompts API handler, the YAML handler and the directory handler. As well as plenty of manual verification testing.


src/tools/auth0/handlers/prompts.ts
```diff
@@ -1,11 +1,96 @@
...
+                  type: 'object',
+                },
+              };
+            }, {}),
+          },
+        };
+      }, {}),
+    },
+  },
+};
```
willvedd, line 66: @sergiughf Similarly to the previous PR's comment, I made sure to validate this. It's especially important for prompts because the schema doesn't match the API responses exactly.

src/context/directory/handlers/prompts.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const promptsSettings = (() => {
+    const promptsSettingsFile = getPromptsSettingsFile(promptsDirectory);
+    return loadJSON(promptsSettingsFile, context.mappings) as PromptSettings;
+  })();
+
+  const customText = (() => {
+    const customTextFile = getCustomTextFile(promptsDirectory);
+    return loadJSON(customTextFile, context.mappings) as AllPromptsByLanguage;
+  })();
```
willvedd, line 40: Need to check and make sure that these files actually exist before `loadJSON`

test/context/directory/prompts.test.ts
```diff
@@ -0,0 +1,146 @@
...
+          login: {
+            buttonText: `${config.AUTH0_KEYWORD_REPLACE_MAPPINGS.BUTTON_TEXT_FRENCH}`,
+            description: 'french login description text',
+          },
+        },
+      },
+    });
+  });
+
+  it('should not dump prompts settings and prompts custom when prompts object is null', async () => {
```
willvedd, line 82: prompts custom text*

test/context/directory/prompts.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    const context = new Context({ AUTH0_INPUT_FILE: dir });
+    context.assets.prompts = null;
+
+    promptsHandler.dump(context);
+    const dumpedFiles = getFiles(promptsDirectory, ['.json']);
+
+    expect(dumpedFiles).to.have.length(0);
+  });
+
+  it('should not dump prompts settings and prompts custom when prompts object is undefined', async () => {
```
willvedd, line 94: prompts custom text*

sergiughf (approved):

PR #534, opened 2022-05-10
Adding `ApiMethodOverride` type for API function overrides

willvedd:
## ✏️ Changes

Adding the `ApiMethodOverride` type to facilitate stronger typing when the API controllers override the default set of CRUD methods. This override is necessary for times when the Node SDK's exported methods for a specific resource don't match the expected `create`, `update` and `delete` naming conventions. Additionally, this override behavior is used to alter the arguments passed into those CRUD functions. For example, the actions controller overrides the create and delete functions in its API controller:

```
functions: {
        create: (action) => this.createAction(action),
        delete: (action) => this.deleteAction(action),
},
```

The override mechanism works either by passing in a string to denote the client method name or a function which will override the function altogether. The underlying mechanism still remains, but the addition of the `ApiMethodOverride` type just makes the contract a bit stronger and allows us to remove a couple `//@ts-ignore`s.

### DIFF
--- a/src/tools/auth0/handlers/actions.ts
+++ b/src/tools/auth0/handlers/actions.ts
@@ -70,7 +70,7 @@ function isActionsDisabled(err) {
 export default class ActionHandler extends DefaultAPIHandler {
   existing: Asset[] | null;
 
-  constructor(options) {
+  constructor(options: DefaultAPIHandler) {
     super({
       ...options,
       type: 'actions',
--- a/src/tools/auth0/handlers/customDomains.ts
+++ b/src/tools/auth0/handlers/customDomains.ts
@@ -33,10 +33,7 @@ export default class CustomDomainsHadnler extends DefaultAPIHandler {
       identifiers: ['domain'],
       stripCreateFields: ['status', 'primary', 'verification'],
       functions: {
-        //@ts-ignore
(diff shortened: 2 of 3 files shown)


src/tools/auth0/handlers/actions.ts
```diff
@@ -70,7 +70,7 @@ function isActionsDisabled(err) {
 export default class ActionHandler extends DefaultAPIHandler {
   existing: Asset[] | null;
 
-  constructor(options) {
+  constructor(options: DefaultAPIHandler) {
```
willvedd, line 73: The actions handler needs this type in the constructor to appropriately evaluate the types of the function overrides it is applying.

src/tools/auth0/handlers/customDomains.ts
```diff
@@ -33,10 +33,7 @@ export default class CustomDomainsHadnler extends DefaultAPIHandler {
       identifiers: ['domain'],
       stripCreateFields: ['status', 'primary', 'verification'],
       functions: {
-        //@ts-ignore
```
willvedd, line 36: Main thing here is the deletion of the `//@ts-ignore`

sergiughf (approved):

Issue #536, opened 2022-05-10
AUTH0_EXCLUDED isn't respect during import


DanPatten:
I am running auth0-cli version 7.12 and included the following:

"AUTH0_EXCLUDED": [ "customDomains" ]
The export works fine and does not include the custom-domains.json file but when importing I get the following error:

Error: <%= "Error: Error: Error parsing JSON from metadata file: tenants\zz\custom-domains\custom-domains.json, because: Unable to load file \zz\custom-domains\custom-domains.json due to Error: ENOENT: no such file or directory, access 'zz\custom-domains\custom-domains.json'" %>

Am I missing something or is this an issue?

willvedd:
Thanks for opening this issue. I was able to recreate the issue and it appears to occur if you have an empty `customDomains` directory in your local configuration directory, regardless of how `AUTH0_EXCLUDED` is set.

So to unblock you in the short term, I advise to just delete the `custom-domains` directory. 

However, this shows that the `AUTH0_EXCLUDED` functionality doesn't apply appropriately during import. Looking at the code again, I think my thinking was that any configuration that was ignored on export would lack configuration and thus be ignored during subsequent import. However, I see now that strategy doesn't cover all possible use cases. It very well could be possible to have configuration in your local directory but you want to ignore certain assets. 

In any case, you have a point here and I'll need to apply a code update to make sure that exclusions apply during import too.

PR #545, opened 2022-05-18
Fix: resource exclusion respected during import

willvedd:
## ✏️ Changes

As pointed out in #536 , resource exclusions configured through the `AUTH0_EXCLUDED` property were only getting applied on export, not import. This issue is not experienced when import occurs after an excluded export because the configurations are omitted completely and ignored on subsequent import. However, in the case of existing resource configurations in the working directory, the exclusions weren't getting applied as expected. This PR corrects this behavior by applying filters to omit excluded resources on file import for both YAML and directory context handlers.

## 🔗 References

- Original Github issue: #536 

## 🎯 Testing

Added tests for both the YAML and directory context handlers to ensure that resource exclusions are getting applied properly during import when resource configurations exist.


src/context/directory/index.ts
```diff
@@ -51,12 +51,17 @@ export default class DirectoryContext {
...
 
-      Object.values(handlers).forEach((handler) => {
-        const parsed = handler.parse(this);
-        Object.entries(parsed).forEach(([k, v]) => {
-          this.assets[k] = v;
+      Object.entries(handlers)
+        .filter(([handlerName]: [AssetTypes, DirectoryHandler<any>]) => {
+          const excludedAssetTypes = this.config.AUTH0_EXCLUDED || [];
+          return !excludedAssetTypes.includes(handlerName);
+        })
```
willvedd, line 58: Only code that has changed in this diff.

src/context/yaml/index.ts
```diff
@@ -69,6 +69,20 @@ export default class YAMLContext {
...
+        if (excludedAssetTypes.includes(key)) return acc;
+
+        return {
+          ...acc,
+          [key]: this.assets[key],
+        };
+      },
+      {}
+    );
+    this.assets = excludedAssetsFiltered;
```
willvedd, line 84: The YAML context handler is a bit different than the directory context handler. It's not excluding the handlers from the batch but instead plucks off the individual values from the main YAML file. The end result is the same, but just the main technical difference between the two context handlers.

Issue #365, opened 2021-07-15
Pages YAML handler type error, expecting buffer received undefined


dejwsz:
Tested on MacOS 11.4
Installed via: npm i -g auth0-deploy-cli

npm version
{
  npm: '7.19.1',
  node: '16.5.0',
  v8: '9.1.269.38-node.20',
  uv: '1.41.1',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.17.1',
  modules: '93',
  nghttp2: '1.43.0',
  napi: '8',
  llhttp: '6.0.2',
  openssl: '1.1.1k',
  cldr: '39.0',
  icu: '69.1',
  tz: '2021a',
  unicode: '13.0'
}

a0deploy export -d -c ./config.json -f yaml -o ./result

2021-07-15T19:19:18.339Z - debug: Start command export
2021-07-15T19:19:18.813Z - info: Loading Auth0 Tenant Data
2021-07-15T19:19:18.813Z - info: Retrieving rules data from Auth0
2021-07-15T19:19:19.031Z - info: Retrieving rulesConfigs data from Auth0
2021-07-15T19:19:19.234Z - info: Retrieving hooks data from Auth0
2021-07-15T19:19:19.414Z - info: Retrieving pages data from Auth0
2021-07-15T19:19:19.844Z - info: Retrieving resourceServers data from Auth0
2021-07-15T19:19:20.023Z - info: Retrieving clients data from Auth0
2021-07-15T19:19:20.215Z - info: Retrieving databases data from Auth0
2021-07-15T19:19:20.413Z - info: Retrieving connections data from Auth0
2021-07-15T19:19:20.606Z - info: Retrieving tenant data from Auth0
2021-07-15T19:19:20.821Z - info: Retrieving emailProvider data from Auth0
2021-07-15T19:19:21.079Z - info: Retrieving emailTemplates data from Auth0
2021-07-15T19:19:27.242Z - info: Retrieving clientGrants data from Auth0
2021-07-15T19:19:27.414Z - info: Retrieving guardianFactors data from Auth0
2021-07-15T19:19:27.586Z - info: Retrieving guardianFactorProviders data from Auth0
2021-07-15T19:19:27.775Z - info: Retrieving guardianFactorTemplates data from Auth0
2021-07-15T19:19:27.972Z - info: Retrieving guardianPolicies data from Auth0
[...]

PR #592, opened 2022-06-24
Prevent crash when writing page templates with YAML handler

willvedd:
## ✏️ Changes

As noted by #365, sometimes the tool crashes when trying to write an HTML template with an undefined value.I do not know why some pages have undefined HTML, and judging by the [GET tenant settings docs](https://auth0.com/docs/api/management/v2#!/Tenants/tenant_settings_route), those page HTML values _should_ be set. However, it is evident that this situation does occur in production. This PR prevents crashes by writing an empty HTML file, essentially forcing the value to an empty string instead of being undefined.

## 🔗 References

Original issue: #365 

## 🎯 Testing

Added test case for pages that return undefined HTML values.

### DIFF
--- a/src/context/yaml/handlers/pages.ts
+++ b/src/context/yaml/handlers/pages.ts
@@ -43,7 +43,7 @@ async function dump(context: YAMLContext): Promise<ParsedPages> {
     // Dump html to file
     const htmlFile = path.join(pagesFolder, `${page.name}.html`);
     log.info(`Writing ${htmlFile}`);
-    fs.writeFileSync(htmlFile, page.html);
+    fs.writeFileSync(htmlFile, page.html || '');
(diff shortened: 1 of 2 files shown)


src/context/yaml/handlers/pages.ts
```diff
@@ -43,7 +43,7 @@ async function dump(context: YAMLContext): Promise<ParsedPages> {
     // Dump html to file
     const htmlFile = path.join(pagesFolder, `${page.name}.html`);
     log.info(`Writing ${htmlFile}`);
-    fs.writeFileSync(htmlFile, page.html);
```
willvedd, line 46: This is the fix.

sergiughf (approved):

PR #599, opened 2022-06-30
Update config.json.example

davidklemme:
## ✏️ Changes

Expanding the example config to explicitly show the region. if the region is not included, the api will return 'unauthorized' and its not immediatly clear why.

No dependencies...

### DIFF
--- a/examples/yaml/config.json.example
+++ b/examples/yaml/config.json.example
@@ -1,5 +1,6 @@
 {
-  "AUTH0_DOMAIN": "<YOUR_TENANT>.<YOUR_REGION>.auth0.com",
+  "AUTH0_DOMAIN": "<YOUR_TENANT>.auth0.com", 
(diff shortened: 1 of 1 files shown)


examples/yaml/config.json.example
```diff
@@ -1,5 +1,6 @@
 {
-  "AUTH0_DOMAIN": "<YOUR_TENANT>.auth0.com",
+  "AUTH0_DOMAIN": "<YOUR_TENANT>.auth0.com", 
+  #include your region if you are not using custom domains. e.g. <YOUR_TENANT>.<YOUR_REGION>.auth0.com
```
willvedd, line 3:
```suggestion
  "AUTH0_DOMAIN": "<YOUR_TENANT>.<YOUR_REGION>.auth0.com", 
```

willvedd (changes requested):
I'm alright with being a bit more verbose with the placeholder but the additional comment is unnecessary IMO.

willvedd (approved):

PR #605, opened 2022-07-16
FIX: Rebuild actions flow deploy on build state issue

anantsharma1310:
## ✏️ Changes

Process actions again when actions are stuck in the build state and the retry limit has been reached.

## 🎯 Testing

Deploy actions multiple times and you'll see the "DEPLOYING STEP:" log which shows intermittent actions failure to deploy as they are stuck in the build state. It starts the rebuilding process with the log "RETRY: Processing actions changes again.". (Current max retry flow attempts: 2)

### DIFF
--- a/src/tools/auth0/handlers/actions.ts
+++ b/src/tools/auth0/handlers/actions.ts
@@ -6,6 +6,7 @@ import { Asset } from '../../../types';
 import { showCompletionScript } from 'yargs';
 
 const MAX_ACTION_DEPLOY_RETRY_ATTEMPTS = 60; // 60 * 2s => 2 min timeout
+const RETRY_FLOW = MAX_ACTION_DEPLOY_RETRY_ATTEMPTS + 2; // Retry rebuilding actions flow 2 times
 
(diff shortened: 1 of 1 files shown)


willvedd:
Is this meant to increase the number of action deployment retries or is it meant to suppress logs? I suppose I'm confused why this code is needed to effectively double the amount of retries. The retry limit is already fairly high at two minutes, I'm really not sure if the action would successfully deploy given two additional minutes. Additionally, not clear to me why you'd reprocess the changes again with `await this.processChanges(assets)`, what is this meant to accomplish?

anantsharma1310:
We are trying to update actions again in case it is stuck in the build state after 2 mins retry, additional retry (2 times)

anantsharma1310:
Hey, we were reprocessing actions in case it is stuck in the build state, another retry we have added is different as we are retrying reprocessing actions and not only redeploying them. This fix works for us but I believe there can be a better and more elegant solution to this problem so I'm closing this pull request.

PR #615, opened 2022-07-27
Enabling HTTP request/response recordings for E2E tests

willvedd:
## ✏️ Changes

Basic E2E tests were included in #611 and #610 but they ran against a live tenant. While this ensures _true_ E2E functionality of the test, it incurs a performance hit and limits the number of test we can run in practice. Additionally, given potential config drift of the target tenant and the unpredictability of network connections, these tests are not expected to be 100% reliable. To improve the predictability, performance and practicality of the tests, we're enabling recording and replaying of HTTP requests. 

This PR includes a decent refactor to allow the recording and replaying, but ultimately the code replaced is much stronger and will ensure better testing overall.

## 🔗 References

Previous E2E testing PRS: #611 #610 

## 🎯 Testing

This PR adds stronger and repeatable E2E tests.


src/context/index.ts
```diff
@@ -64,6 +64,7 @@ export const setupContext = async (config: Config): Promise<DirectoryContext | Y
   })(config);
 
   const accessToken = await (async (): Promise<string> => {
+    if (config.AUTH0_DOMAIN === 'deploy-cli-dev.eu.auth0.com') return '[REDACTED]';
```
willvedd, line 67: This hacky bit is required because we're redacting the access token when we post-process the Nock recordings. However, Nock will try to proceed through the normal authentication flow and be unable to match the non-redacted authentication requests with the redacted recordings. By adding this line of code here, we're immediately bypassing the access token generation because it's not needed by the recordings.

.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     steps:
       - checkout
       - run: npm ci
-      - run: npm run test:e2e:node-module
+      - run: AUTH0_HTTP_RECORDINGS="on" npm run test:e2e:node-module
```
willvedd, line 16: Enabling the usage of recordings in CI.

package.json
```diff
@@ -10,8 +10,8 @@
     "lint:fix": "eslint --fix --ignore-path .eslintignore --ignore-pattern webpack . && kacl lint",
     "lint": "eslint --ignore-path .eslintignore --ignore-pattern webpack . && kacl lint",
     "format": "npx prettier --write .",
-    "test": "ts-mocha -p tsconfig.json --recursive 'test/**/*.test*' --exclude test/e2e/* --timeout 5000",
-    "test:e2e:node-module": "ts-node test/e2e/e2e-node-module.ts",
+    "test": "ts-mocha -p tsconfig.json --recursive 'test/**/*.test*' --exclude 'test/e2e/*' --timeout 5000",
```
willvedd, line 13: Needed to fix this exclude pattern, otherwise the E2E tests would get run when running `npm run test`.

test/e2e/e2e.test.ts
```diff
@@ -0,0 +1,87 @@
+import { expect } from 'chai';
+import path from 'path';
+import { getFiles, existsMustBeDir } from '../../src/utils';
+import { setupRecording, testNameToWorkingDirectory } from './e2e-utils';
+import { dump, deploy } from '../../src';
+
+const shouldUseRecordings = process.env['AUTH0_HTTP_RECORDINGS'] === 'on';
+const AUTH0_DOMAIN = shouldUseRecordings
```
willvedd, line 8: The value of `AUTH0_HTTP_RECORDINGS` env var dictates wether the recordings are used or not. And if not, set the `AUTH0_DOMAIN` to the fake `deploy-cli-dev.eu.auth0.com` domain.

test/e2e/e2e-utils.ts
```diff
@@ -0,0 +1,96 @@
...
+export function testNameToWorkingDirectory(testName = ''): string {
+  const directoryName = testName.replaceAll(' ', '-');
+  return path.join('./local/recorded', directoryName);
+}
+
+export function sanitizeRecording(recording: Definition): Definition {
+  const sanitizedRecording = recording;
+
+  //@ts-ignore because the `rawHeaders` property does actually exist
+  sanitizedRecording.rawHeaders = [];
```
willvedd, line 43: For some reason the type declarations in Nock is wrong. The `Definition` type should have the `rawHeaders` property.

test/e2e/e2e-utils.ts
```diff
@@ -0,0 +1,96 @@
...
+  sanitizedRecording.scope = 'https://deploy-cli-dev.eu.auth0.com:443';
+  sanitizedRecording.body = sanitizeObject(recording.body, ['client_secret']);
+  sanitizedRecording.response = sanitizeObject(recording.response, [
```
willvedd, line 46: Where we are both sanitizing the request and the response.

test/e2e/e2e-utils.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  return obj;
+};
+
+//Nock Config
+import { back as nockBack } from 'nock';
+nockBack.setMode(process.env['AUTH0_HTTP_RECORDINGS'] === 'on' ? 'lockdown' : 'wild');
+nockBack.fixtures = path.join(__dirname, 'recordings');
+
+export function setupRecording(testName) {
```
willvedd, line 85: Set up this helper function that configures the nock recording so that the implementation in the tests is a one-liner that is consistent for all tests.

sergiught (approved):

Issue #607, opened 2022-07-21
Applying a dumped passwordless email configuration with custom email template is not working


lordgordon:
### Description

After configuring the passwordless email flow from Auth0 Web UI, the deploy cli is unable to properly handle a custom email template.

### Reproduction

1. configure passwordless email: https://manage.auth0.com/dashboard/***/***/connections/passwordless
2. add a custom body and save
![image](https://user-images.githubusercontent.com/7745293/180222915-29dcd726-5217-4fa8-88b5-6472cc18b659.png)
3. run the dump `a0deploy dump --export_ids=true --env -c config-qa.json -f yaml -o auth0-dump-qa`
4. this adds the body in `./connections/email.html` (the html is correctly dumped) and the new settings (see below)
5. run apply `a0deploy import --env -c config-qa.json -i qa/tenant.yaml`
6. now from Auth0 UI the body is incorrectly overwritten: 
![image](https://user-images.githubusercontent.com/7745293/180224325-f5b4af91-9112-4872-815c-6d24cd103acc.png)
8. and the actual email is just the `./connections/email.html` text without my custom template and login url

Both `dump` and `apply` run without errors or warnings.

Here my current `config.json`:
```json
{
  "AUTH0_DOMAIN": "***.auth0.com",
  "AUTH0_CLIENT_ID": "***",
  "AUTH0_ALLOW_DELETE": true,
  "EXCLUDED_PROPS": {
    "connections": ["options.client_secret"],
    "emailProvider": ["credentials.smtp_pass"]
  },
  "AUTH0_EXCLUDED": ["themes"]
}

```

Here the relevant diff of the new sections added by the dump for the passwordless settings:
```yaml
connections:
 - name: email
    strategy: email
    id: ***
    enabled_clients:
      - my-app
    is_domain_connection: false
    options:
      name: email
      totp:
        length: 6
        time_step: 900
      email:
        body: ./email.html
        from: MyApp <login@myappcom>
        syntax: liquid
        subject: Welcome in MyApp!
      disable_signup: false
      brute_force_protection: true
```

### Environment

- auth0-deploy-cli version: `7.14.1`

lordgordon:
I tried to workaround this by adding `"options.email.body"` to `EXCLUDED_PROPS` for `connections`:

```json
{
  "AUTH0_DOMAIN": "***",
  "AUTH0_CLIENT_ID": "***",
  "AUTH0_ALLOW_DELETE": true,
[...]

PR #617, opened 2022-07-29
Fix: Throw error if passwordless email template is missing

willvedd:
## ✏️ Changes

As discovered in #607, if a passwordless email connection's  path to the email template is missing for whatever reason, the body of the email template gets replaced with that path. 

Consider the following configuration:

```yaml
connections:
  - name: email
    strategy: email
    enabled_clients:
      - 3glCfiBX6R8pZBPIaI89F1u5McZ3zoHB
      - PewQpGx73sHcsV6ufVNZNrV4GDlDDm71
    is_domain_connection: false
    options:
      name: email
      email:
        body: ./this-path-wont-resolve/email.html # NOTE THIS PATH
        from: '{{ application.name }} <root@auth0.com>'
        syntax: liquid
        subject: Welcome to {{ application.name }}
      disable_signup: false
      brute_force_protection: true
```

If the above is applied, the process will succeed despite the email template being missing. Further, the body of the template will be literally replaced with the string `./this-path-wont-resolve/email.html`. 

This PR introduces safeguards that will prevent this unintentional destruction from occurring. It works by throwing and thus halting the deployment process from proceeding to the process stage (API requests). This is implemented for both the YAML and directory formats.

## 🔗 References

Original Github issue: #607 

## 🎯 Testing

Added unit tests for both cases.

### DIFF
--- a/src/context/yaml/handlers/connections.ts
+++ b/src/context/yaml/handlers/connections.ts
@@ -31,9 +31,12 @@ async function parse(context: YAMLContext): Promise<ParsedConnections> {
           ensureProp(connection, 'options.email.body');
           const htmlFileName = path.join(connectionsFolder, connection.options.email.body);
 
-          if (isFile(htmlFileName)) {
-            connection.options.email.body = context.loadFile(htmlFileName);
+          if (!isFile(htmlFileName)) {
+            const missingTemplateErrorMessage = `Passwordless email template purportedly located at ${htmlFileName} does not exist for connection. Ensure the existence of this file to proceed with deployment.`;
+            log.error(missingTemplateErrorMessage);
+            throw new Error(missingTemplateErrorMessage);
           }
+          connection.options.email.body = context.loadFile(htmlFileName);
         }
 
         return connection;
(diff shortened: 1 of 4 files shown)


src/context/yaml/handlers/connections.ts
```diff
@@ -31,9 +31,12 @@ async function parse(context: YAMLContext): Promise<ParsedConnections> {
           ensureProp(connection, 'options.email.body');
           const htmlFileName = path.join(connectionsFolder, connection.options.email.body);
 
-          if (isFile(htmlFileName)) {
-            connection.options.email.body = context.loadFile(htmlFileName);
+          if (!isFile(htmlFileName)) {
+            const missingTemplateErrorMessage = `Passwordless email template purportedly located at ${htmlFileName} does not exist for connection. Ensure the existence of this file to proceed with deployment.`;
+            log.error(missingTemplateErrorMessage);
```
willvedd, line 36: For some reason, the YAML handler does not surface the thrown error in a clean way, unlike the directory handler that does surface the message. This is meant to explain this subtle inconsistency. Fixing it outside the scope of this ticket IMO.

sergiught (approved):

sdmcintyre2:
This change has broken inline email templates. Was that intended?

PR #626, opened 2022-08-11
Remove unnecessary dependencies and move others to dev-dependencies

### LINKED ISSUE #625: Dependency on ts-node
### Description

Updating from `7.14.1` to `7.14.2` added `ts-node` as a dependency. It would be better if `auth0-deploy-cli` was bundled into a pure JS module rather than relying on Typescript at runtime. Or maybe was this a misplaced devDependency ?

### Reproduction

Add the latest version of `auth0-deploy-cli` as dependency. Notice that `ts-node` was installed.

### Environment

> Please provide the following:

- **Version of this library used:** `7.14.2`
- **Version of the platform or framework used, if applicable:** `v16.16.0`
- **Other relevant versions (language, server software, OS, browser):** N.A.
- **Other modules/plugins/libraries that might be involved:** `ts-node`

jmac105:
## ✏️ Changes

Removing dependencies which are not unused, and moving other packages from dependencies into dev-dependencies such as test assertion libraries. 
Resolves #625 

## 🎯 Testing

No tests added or removed, tests are all passing locally. 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,30 +77,30 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.17.7",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.17.7.tgz",
-      "integrity": "sha512-p8pdE6j0a29TNGebNm7NzYZWB3xVZJBZ7XGs42uAKzQo8VQ3F0By/cQCtUEABwIqw5zo6WA4NbmxsfzADzMKnQ==",
+      "version": "7.18.8",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.18.8.tgz",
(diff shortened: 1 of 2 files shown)


willvedd (dismissed):
Ha! You beat me to it, I already had this exact set of changes ready to push. Anyways, appreciate you addressing this.

jmac105:
Not sure why e2e tests are failing, as far as I can tell it does not seem to be related to my changes?

willvedd:
@jmac105 Yeah, I think it has to do with this being a forked PR. I'm looking into it now and will re-run locally but anticipate your changes to be valid. Once I confirm, I'll go ahead and merge.

willvedd (approved):
In addition to your changes, I removed cross-env dependency as it isn't being used anymore. Was able to confirm that e2e test failures are the result of secrets not being passed through for forked PRs, will need to address that sometime in the near future.

PR #634, opened 2022-08-23
DXCDT-208: Analytics headers

willvedd:
## ✏️ Changes

Some of the downstream data pipelines and analytics platforms struggle with parsing data from the user-agent header. This PR separates these data attributes into their own specific header. In theory, this _should_ be possible through modifying the correct data pipelines, but that is a much straightforward process. These changes may be reverted when the data pipeline is able to perform this type of parsing.

These are data properties that will help us better understand the usage of this tool to then be able to make data-driven decisions about all aspects of this project.

In addition to the HTTP header separation, this PR also adds a random and arbitrary UUID to mark unique "sessions" this is just an extra data point that will inform how often the tool is invoked.

## 🎯 Testing

This needs to be tested by making a test request and detecting the presence (or absence) of these values in data pipeline. **Still a draft until confirmed**.

### DIFF
--- a/src/context/index.ts
+++ b/src/context/index.ts
@@ -1,4 +1,5 @@
 import path from 'path';
+import { randomUUID } from 'crypto';
 import { AuthenticationClient, ManagementClient } from 'auth0';
 import YAMLContext from './yaml';
 import DirectoryContext from './directory';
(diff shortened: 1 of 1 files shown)


willvedd:
Closing this one out due to a few things learned during metric collection discovery that will change the approach here. Will re-up something similar in the future.

Issue #648, opened 2022-09-07
Inconsistency in branding -> (colors, logo_url) between a0deploy versions and formats


joshw:
### Description

We have been using version 7.9.0, and when I went to update to 7.14.3, i noticed some changes in the yaml format output of export.   I tried several versions, and it seems like this changed between 7.10.0 and 7.11.0.

Specifically, in version 7.11.0 and higher, the "branding" section is missing the colors and logo_url properties in the yaml output format.  They do seem to be present in the directory format.

I think these fields were accidentally removed from the yaml format, but even if this change was deliberate, it seems like the two output formats should agree.

> Provide a clear and concise description of the issue, including what you expected to happen.

### Reproduction

In 7.10.0, in the yaml format, tenant.yaml contains
```yaml
branding:
  colors:
    primary: '#f55e05'
    page_background: '#000000'
  logo_url: <my url>
  templates: []
```
and in directory mode, branding/branding.json contains
```json
{
  "colors": {
    "primary": "#f55e05",
    "page_background": "#000000"
  },
  "logo_url": "<my url>"
}
```

In newer versions, however, this is different.    In 7.14.3, I get this instead:

yaml mode:
```yaml
branding:
  templates: []
```
directory mode:
```json
{
  "colors": {
    "primary": "#f55e05",
    "page_background": "#000000"
  },
  "logo_url": "<url>"
}
```

willvedd:
Yep, this is certainly a bug. I've identified that somewhere between 7.9.0 and current, the non-template branding settings are omitted for the YAML configuration format. Will look to address this very soon.

willvedd:
Thanks for opening this issue. It has been fixed with [7.15.1](https://github.com/auth0/auth0-deploy-cli/releases/tag/v7.15.1).

PR #666, opened 2022-10-13
Returning all branding settings for YAML branding handler

willvedd:
## ✏️ Changes

In response to #648, which highlighted that there was missing branding configurations when using the YAML format. For example, when exporting branding settings in recent versions, there is a clear disprepancy in exported config:

**YAML:**
```
branding:
  templates: []
```

**Directory:**
```
{
  "colors": {
    "primary": "#f55e05",
    "page_background": "#000000"
  },
  "logo_url": "<url>"
}
```

This PR addresses the missing branding properties when using the YAML format. 

## 🔗 References

Original issue: #648 

## 🎯 Testing

One of the reasons this made it's way into production is because the existing test for the branding YAML handler _only_ checked for template-related configuration, completely ignoring all other possible branding settings. This PR introduces more comprehensive testing to assert that all branding config gets dumped.

### DIFF
--- a/src/context/yaml/handlers/branding.ts
+++ b/src/context/yaml/handlers/branding.ts
@@ -19,14 +19,6 @@ type ParsedBranding = ParsedAsset<
 >;
 
 async function parse(context: YAMLContext): Promise<ParsedBranding> {
-  // Load the HTML file for each page
-  if (!context.assets.branding)
-    return {
-      branding: {
-        templates: [],
-      },
-    };
-
   if (!context.assets.branding) return { branding: null };
 
   const {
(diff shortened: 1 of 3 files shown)


src/context/yaml/handlers/branding.ts
```diff
@@ -19,14 +19,6 @@ type ParsedBranding = ParsedAsset<
...
 
 async function parse(context: YAMLContext): Promise<ParsedBranding> {
-  // Load the HTML file for each page
-  if (!context.assets.branding)
-    return {
-      branding: {
-        templates: [],
-      },
-    };
-
```
willvedd, line 29: Removed because a more appropriate early return is done on the following line.

src/context/yaml/handlers/branding.ts
```diff
@@ -91,7 +83,7 @@ async function dump(context: YAMLContext): Promise<ParsedBranding> {
     });
   }
 
-  return { branding: { templates } };
+  return { branding: { templates, ...branding } };
```
willvedd, line 86: What ultimately fixes the original issue.

sergiught (approved):

Issue #656, opened 2022-09-20
blank branding.logo_url is exported if I don't set anything to Company Logo, but this configuration causes failure while importing


shibayu36:
### Description
I don't set anything to Company Logo in "Branding > Universal Login", and I execute `a0deploy export`, then it exports blank branding.logo_url like following

branding/branding.json
```json
{
  "colors": {
    "page_background": "#000000",
    "primary": "#ea5323"
  },
  "logo_url": ""
}
```

After exporting, I execute `a0deploy import` using these exported settings, then this command failed with error.

```
2022-09-20T14:07:14.738Z - error: Problem running command import during stage processChanges when processing type branding
2022-09-20T14:07:14.738Z - error: Payload validation error: 'Object didn't pass validation for format strict-https-uri-or-null: ' on property logo_url (URL for the logo. Must use HTTPS).
```

### Reproduction

- Set blank to Company Logo in "Branding > Universal Login"
    - <img width="502" alt="image" src="https://user-images.githubusercontent.com/1132489/191283388-5a070d0f-cc88-4e59-958a-c661501d44e4.png">
- Execute `a0deploy export`
- Execute `a0deploy import`

### Environment

> Please provide the following:

- **Version of this library used:** 7.14.3
- **Version of the platform or framework used, if applicable:** Private Cloud v202235.81.6
- **Other relevant versions (language, server software, OS, browser):**
- **Other modules/plugins/libraries that might be involved:**

willvedd:
This has been addressed with [7.15.1](https://github.com/auth0/auth0-deploy-cli/releases/tag/v7.15.1).

shibayu36:
> Can you think of any reasons why your tenant would be returning that value? Is it a brand new tenant? Private cloud? Are you using classic login?

I use Private Cloud and classic login.  I don't need to set `logo_url` because I use a full customized HTML for classic login.

> I've tried again, and I still can't recreate this issue, for either YAML or directory formats

hmm...
Before, I found that some fields were not exported but once I pressed a save button in the Dashboard then those fields appeared with blank.  So you may reproduce it if you press the save button in "Branding > Universal Login" with changing a primary color and keeping logo_url blank.

shibayu36:
I think it is a problem on the API side.  So I also created a ticket to support.auth0.com

Issue #226, opened 2020-04-17
Blank logo_url causing import to fail


mogusbi-motech:
### Description

Having run an export for my tenant, I get the following for the branding section of the output `yaml`;

```yaml
branding:
  colors:
    page_background: '#161616'
    primary: '#2e9dc8'
  logo_url: ''
```

However, when I then go to import that I get the following error;

`Payload validation error: 'Object didn't pass validation for format strict-https-uri-or-null: ' on property logo_url (URL for the logo. Must use HTTPS).`

I can fix this by either bumping off the `logo_url` property or setting it to null but when I go to export my settings again it automatically gets added again.

### Reproduction

- Edit tenant settings on auth0.com but leave the logo field blank
- Export the config using `a0deploy export -c config.json -f yaml -o ./src`
- Import the config without making any changes using `a0deploy import -c config.json -i ./src/tenant.yaml`

### Environment

> Please provide the following:

- **Version of this library used:** 4.2.1
- **Version of the platform or framework used, if applicable:**
- **Other relevant versions (language, server software, OS, browser):** OS X 10.15.4
- **Other modules/plugins/libraries that might be involved:**

willvedd:
Confirming that this issue still active. Reproducible if you export and import configuration from a fresh tenant.

PR #667, opened 2022-10-13
DXCDT-240: Prevent empty `logo_url` update payload inclusion

willvedd:
## ✏️ Changes

Under certain and unknown, conditions, an empty string `logo_url` is returned when fetching branding settings. While I have not been able to reproduce the export of a blank `logo_url`, is has been reported by #226 and #656, suggesting that real folks actually experience this. The proposed validation is simple enough and doesn't produce significant tech debt. 

## 🔗 References

Original issues: #226, #656

## 🎯 Testing

Added a test case for `logo_url` being omitted from update payload. 

Additionally, added an unrelated but important test around the prevention of sending an API request if branding settings is an empty object.

### DIFF
--- a/test/tools/auth0/handlers/branding.tests.js
+++ b/test/tools/auth0/handlers/branding.tests.js
@@ -161,6 +161,48 @@ describe('#branding handler', () => {
 
       const auth0 = {
         branding: {
+          updateSettings: (_params, data) => {
+            expect(data).to.deep.equal({
+              colors: {
+                primary: '#F8F8F2',
+                page_background: '#112',
+              },
+              font: {
+                url: 'https://mycompany.org/font/myfont.ttf',
+              },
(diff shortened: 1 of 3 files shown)


test/tools/auth0/handlers/branding.tests.js
```diff
@@ -156,6 +156,74 @@ describe('#branding handler', () => {
...
+              url: 'https://mycompany.org/font/myfont.ttf',
+            },
+          },
+        },
+      ]);
+
+      expect(wasUpdateCalled).to.equal(true);
+    });
+
+    it('should not send updateSettings request if empty object passed', async () => {
```
willvedd, line 201: This test is unrelated, but still a valid case nonetheless.

sergiught (approved):

Issue #654, opened 2022-09-16
Bug: export failure when trying dump undefined `page.html`


sandersiim:
### Description

Running `export` command fails if there is a `page` with undefined `html`.

### Reproduction

When using export functionality with `--format directory --export_ids --debug`, the following error can happen when there is a Page without html:

```
2022-09-16T08:26:10.203Z - debug: TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined
    at Object.writeFileSync (node:fs:2200:5)
    at /Users/sander.siim/auth0-fetcher/node_modules/auth0-deploy-cli/lib/context/directory/handlers/pages.js:69:36
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/Users/sander.siim/auth0-fetcher/node_modules/auth0-deploy-cli/lib/context/directory/handlers/pages.js:63:15)
    at Generator.next (<anonymous>)
    at /Users/sander.siim/auth0-fetcher/node_modules/auth0-deploy-cli/lib/context/directory/handlers/pages.js:8:71
```

Issue is caused by incorrect Boolean logic here:
```
diff --git a/src/context/directory/handlers/pages.ts b/src/context/directory/handlers/pages.ts
index a346980..4117dcf 100644
--- a/src/context/directory/handlers/pages.ts
+++ b/src/context/directory/handlers/pages.ts
@@ -69,7 +69,7 @@ async function dump(context: DirectoryContext): Promise<void> {
   pages.forEach((page) => {
     var metadata = { ...page };

-    if (page.name !== 'error_page' || page.html !== undefined) {
+    if (page.name !== 'error_page' && page.html !== undefined) {
       // Dump template html to file
       const htmlFile = path.join(pagesFolder, `${page.name}.html`);
       log.info(`Writing ${htmlFile}`);
```

### Environment

> Please provide the following:

- **Version of this library used:** 7.14.3
- **Version of the platform or framework used, if applicable:** any
- **Other relevant versions (language, server software, OS, browser):**
- **Other modules/plugins/libraries that might be involved:**

willvedd:
You're definitely correct about the source of the error. This project has been suffering from template issues for while now, see: #365  #632 #417. The main problem is that the data is inconsistent between templates. I'm going to try and take a stab at applying more descriptive types anywhere we use `fs.writeFileSync()` to avoid writing undefined to disk.

[...]

PR #671, opened 2022-10-18
DXCDT-237: Fixing undefined HTML page template writing

willvedd:
### 🔧 Changes

Under certain conditions, the `html` property of template page (error page, MFA guardian, login, etc) will be undefined. The Deploy CLI will simply attempt to write that value and will incur the following error:

```sh
TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined
```

This can be observed in the Github issues #654 and #365. Previously, this situation occurred because the logic attempted to predict the existence of properties based on the page type alone. Instead, this PR will largely disregard the page type and just observe the existence of `html` property. If it does, write it to disk, otherwise continue.

To help this process, I added a `Page` type that models the uncertainty of the page template properties. Further, I added the `@types/fs-extra` package which will emit a compilation error if a potentially undefined value may be written to a file. 

### 📚 References

Related Github issue: #654, #365

### 🔬 Testing

Added tests for one or more page templates lacking the `html` property.

### 📝 Checklist

### DIFF
--- a/src/context/directory/handlers/databases.ts
+++ b/src/context/directory/handlers/databases.ts
@@ -125,13 +125,14 @@ async function dump(context: DirectoryContext): Promise<void> {
         // customScripts option only written if there are scripts
         ...(database.options.customScripts && {
           customScripts: Object.entries(database.options.customScripts)
-            //@ts-ignore
+            //@ts-ignore because we'll fix this in subsequent PR
             .sort(sortCustomScripts)
             .reduce((scripts, [name, script]) => {
               // Dump custom script to file
               const scriptName = sanitize(`${name}.js`);
               const scriptFile = path.join(dbFolder, scriptName);
               log.info(`Writing ${scriptFile}`);
+              //@ts-ignore because we'll fix this in subsequent PR
               fs.writeFileSync(scriptFile, script);
               scripts[name] = `./${scriptName}`;
               return scripts;
--- a/src/context/directory/handlers/pages.ts
+++ b/src/context/directory/handlers/pages.ts
@@ -71,7 +71,7 @@ async function dump(context: DirectoryContext): Promise<void> {
     if (page.html !== undefined) {
       const htmlFile = path.join(pagesFolder, `${page.name}.html`);
       log.info(`Writing ${htmlFile}`);
-      fs.writeFileSync(htmlFile || '', page.html);
+      fs.writeFileSync(htmlFile, page.html);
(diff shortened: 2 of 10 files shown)


src/context/directory/handlers/databases.ts
```diff
@@ -125,13 +125,14 @@ async function dump(context: DirectoryContext): Promise<void> {
...
           customScripts: Object.entries(database.options.customScripts)
-            //@ts-ignore
+            //@ts-ignore because we'll fix this in subsequent PR
             .sort(sortCustomScripts)
             .reduce((scripts, [name, script]) => {
               // Dump custom script to file
               const scriptName = sanitize(`${name}.js`);
               const scriptFile = path.join(dbFolder, scriptName);
               log.info(`Writing ${scriptFile}`);
+              //@ts-ignore because we'll fix this in subsequent PR
```
willvedd, line 135: With the addition of the `@types/fs-extra` package, we found a couple of compilation errors that are identical to the one we're trying to fix with pages but with the database resource. We'll fix those later.

src/context/directory/handlers/pages.ts
```diff
@@ -60,24 +61,21 @@ function parse(context: DirectoryContext): ParsedPages {
 async function dump(context: DirectoryContext): Promise<void> {
   const pages = context.assets.pages;
 
-  if (!pages) return; // Skip, nothing to dump
+  if (!pages) return;
 
-  // Create Pages folder
```
willvedd, line 65: Ridding some unhelpful comments.

src/context/directory/handlers/pages.ts
```diff
@@ -60,24 +61,21 @@ function parse(context: DirectoryContext): ParsedPages {
...
   const pagesFolder = path.join(context.filePath, constants.PAGES_DIRECTORY);
   fs.ensureDirSync(pagesFolder);
 
   pages.forEach((page) => {
-    var metadata = { ...page };
+    const metadata = { ...page };
 
-    if (page.name !== 'error_page' || page.html !== undefined) {
-      // Dump template html to file
+    if (page.html !== undefined) {
```
willvedd, line 72: This is ultimately what fixes the issue. We don't care about the type of the page, but we do care if `page.html` is defined or not.

sergiught (approved):

PR #754, opened 2023-02-23
DXCDT-370: Check prerequisites for keyword preservation

willvedd:
### 🔧 Changes

In order for keyword preservation to work, there are two key prerequisites that must be met:
- Non-empty keyword replace mappings
- Non-empty local resource configuration files

If the above criteria are not met, it suggests an erroneous condition and that data loss could occur. This PR introduces checks to ensure that these prerequisites are met before attempting to preserve keywords. Otherwise, an exception will be thrown. It is debatable wether or not an exception is a bit harsh, but for this first iteration I'm proposing to stay on the safer side and stop the usage of keyword preservation if it is doomed to work.

### 🔬 Testing

Added E2E tests for both checks.

### 📝 Checklist


src/context/index.ts
```diff
@@ -25,7 +26,10 @@ const nonPrimitiveProps: (keyof Config)[] = [
   'INCLUDED_PROPS',
 ];
 
-export const setupContext = async (config: Config): Promise<DirectoryContext | YAMLContext> => {
+export const setupContext = async (
+  config: Config,
+  command: 'import' | 'export'
```
willvedd, line 31: It was necessary for me to introduce this command argument to exclude the prerequisite checks if an import was being run.

sergiught (approved):

PR #756, opened 2023-02-28
DXCDT-382: Standardizing Auth0 Resource Identifiers (1/x)

willvedd:
### 🔧 Changes

The Deploy CLI has the concept of "identifiers" which is a set of strings for each Auth0 resource. It contains non-ID unique properties that are used to positively identify a specific entity across tenants. In most cases, the identifier is `name`, but some resources do not have `name` and instead need to rely on another resource-specific identifier. Example: custom domains uses the `domain` field. 

Before keyword preservation can operate on all Auth0 resources using these identifiers, we need to clean-up and make these fields consistently implemented across the codebase.

In theory, no functional changes should have been made. And if there are, they are not regressions. Those changes will be noted below.

### 🔬 Testing

In theory, no functional changes should have been made. All tests pass.

### 📝 Checklist


src/tools/auth0/handlers/emailTemplates.ts
```diff
@@ -23,7 +23,7 @@ export default class EmailTemplateHandler extends DefaultHandler {
     super({
       ...options,
       type: 'emailTemplates',
-      id: 'template',
+      identifiers: ['template'],
```
willvedd, line 26: The keyword preservation algorithm will be leveraging the `identifiers` field for all handlers. This is why the `id` field was removed in favor of `identifiers` here. Because email templates does not use the normal diff'ing algorithm as most other resource in the codebase, this will have no negative effects.

src/tools/auth0/handlers/organizations.ts
```diff
@@ -253,7 +252,7 @@ export default class OrganizationsHandler extends DefaultHandler {
       handler: this,
       assets: organizations,
       existing,
-      identifiers: ['id', 'name'],
+      identifiers: this.identifiers,
```
willvedd, line 255: `['id', 'name']` is the default value, no need to define specifically.

src/tools/auth0/handlers/pages.ts
```diff
@@ -41,6 +41,7 @@ export default class PagesHandler extends DefaultHandler {
     super({
       ...options,
       type: 'pages',
+      identifiers: ['name'],
```
willvedd, line 44: By adding an identifier field here, we are able to enable keyword preservation for pages.

src/tools/auth0/handlers/rules.ts
```diff
@@ -91,7 +91,7 @@ export default class RulesHandler extends DefaultHandler {
       handler: this,
       assets: rules,
       existing,
-      identifiers: ['id', 'name'],
+      identifiers: this.identifiers,
```
willvedd, line 94: `['id', 'name']` is the default value, no need to define specifically.

src/tools/auth0/handlers/roles.ts
```diff
@@ -190,7 +189,7 @@ export default class RolesHandler extends DefaultHandler {
       handler: this,
       assets: roles,
       existing,
-      identifiers: ['id', 'name'],
+      identifiers: this.identifiers,
```
willvedd, line 192: `['id', 'name']` is the default value, no need to define specifically.

sergiught (approved):

Issue #688, opened 2022-11-04
Proposal: Keyword Preservation During Export


willvedd:
### Checklist

- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

-

### Describe the ideal solution

-

### Alternatives and current workarounds

-

### Additional context

-

hayashi-ay:
Thanks for the proposal but I personally don't need these features. They only complicate the cli.

The cause of the problem of keyword replacement is that the import and export configuration files are the same, but in our operation, they are separate files, so there is no problem.

The exported config files are also used as a backup in case the tenant's configuration changes unexpectedly during operation via Auth0 dashboard or the cli.

Of course, it is necessary to manually reflect the changes to the import configuration files, but there are not that many lines to change and no more worrying about whether REMOTE or LOCAL should be prioritized.

jcerjak:
Thanks for the writeup, @willvedd. In general, I think the proposal is sensible. I have some comments/proposals, please see below.

## Comments

> There are a few options available for addressing instances of the reconciliation problem. Certainly, the Deploy CLI will attempt to preserve keywords when the situation is trivial

I think it would be good to define what situations are trivial, and what are considered a reconciliation problem.

Given your comment about magic:
> The name is intentional to convey the possibility of problematic reconciliations, making it clear that the functionality is not performing any magic.

I would initially go for the simplest and safest solution, and not try to do any merge-like reconciliations at all. I propose it would work like this:
* If remote value matches local value (with keyword replacements applied) EXACTLY, keep the local value (with keywords).
* Otherwise keep the remote value.

This would allow:
* Basic export/import roundtrip flow to work
* Easily detect any remote changes (e.g. by doing a git diff), and manually reconcile them by editing local config.

For workflows that rely on frequent remote updates, a more complex/magic approach might be needed. But that can be added later, and shouldn't be the default.

> Prefer Local Values During Reconciliation
> This behavior will allow the preservation of all keyword markers in local configuration and keep most changes on remote. However, this opens the possibility of some remote changes being unkept when local values are preferred. These instances will be communicated to the user through logging outputs.

[...]

PR #760, opened 2023-03-02
DXCDT-388: Enabling preservation of array replace markers in YAML

willvedd:
### 🔧 Changes

One of the final pieces for delivery a MVP deliverable for keyword preservation on export is enabling the preservation of array replace keywords (ex: `@@LOGOUT_URLS@@`) in YAML configuration formats. The hurdle with this limitation can be broken down into two major problems:

**Problem Statement 1:**
When using array replace syntax in a YAML configuration file, the markers can be expressed as-is without quotes, which is invalid YAML. There is a built-in YAML to JSON conversion step that will fail. Example of an error-causing configuration:

```yaml
tenant:
  allowed_logout_urls: @@LOGOUT_URLS@@
```

**Solution:**
Introduce an intermediate step before the YAML parsing that looks for unquoted array replace markers and upon finding them, wrap in quotes. Array replace markers that are already wrapped in quotes should be left as-is. 

--- 

**Problem Statement 2:**
When using array replace syntax in a YAML configuration file, the markers can be expressed as-is without quotes, wrapped with double quotes or wrapped in single quotes. While technically always true, the keyword replacement process preserves these array values wrapped in single-quotes.

For example, all of the following are valid configuration files:
```yaml
tenant:
  allowed_logout_urls: @@LOGOUT_URLS@@
```
```yaml
tenant:
  allowed_logout_urls: '@@LOGOUT_URLS@@'
```
```yaml
tenant:
  allowed_logout_urls: "@@LOGOUT_URLS@@"
```

**Solution:**

Appropriately handle each scenario during the keyword replacement step. In practice, this means introducing a third regex replacement step for single-quoted strings in addition to the double-quoted and unquoted array replace markers.

---

Together, the combination of these solutions enabled a seamless experience for users needing to preserve array replace keywords in the YAML format.

### 📚 References

Related RFC: #688 

### 🔬 Testing

Added new unit tests for any functions that were altered. Additionally, added a couple assertions to E2E tests. Finally, did quite a bit of manual testing for another layer of verification.

### 📝 Checklist


src/context/yaml/handlers/actions.ts
```diff
@@ -86,7 +86,8 @@ async function dump(context: YAMLContext): Promise<ParsedActions> {
       runtime: action.runtime,
       dependencies: action.dependencies || [],
       status: action.status,
-      secrets: mapSecrets(action.secrets || []),
+      secrets:
+        typeof action.secrets === 'string' ? action.secrets : mapSecrets(action.secrets || []), //Enables keyword preservation to operate on action secrets
```
willvedd, line 90: Most array preservation will work as-is for all YAML handlers, but there is this one instance where a string value for `action.secrets` would trigger a runtime error when applying a map function.

src/tools/utils.ts
```diff
@@ -8,9 +8,11 @@ import constants from './constants';
 
 export const keywordReplaceArrayRegExp = (key) => {
   const pattern = `@@${key}@@`;
-  const patternWithQuotes = `"${pattern}"`;
+  //YAML format supports both single and double quotes for strings
+  const patternWithSingleQuotes = `'${pattern}'`;
+  const patternWithDoubleQuotes = `"${pattern}"`;
 
-  return new RegExp(`${patternWithQuotes}|${pattern}`, 'g');
+  return new RegExp(`${patternWithSingleQuotes}|${patternWithDoubleQuotes}|${pattern}`, 'g');
```
willvedd, line 15: Array replace markers could be wrapped in single quotes, double quotes or not wrapped at all. When performing this replacement, it is necessary to check for all three.

src/tools/utils.ts
```diff
@@ -45,6 +47,20 @@ export function keywordReplace(input: string, mappings: KeywordMappings): string
   return input;
 }
 
+// wrapArrayReplaceMarkersInStrings will wrap array replacement markers in quotes.
+// This is necessary for YAML format in the context of keyword replacement
+// to preserve the keyword markers while also maintaining valid YAML syntax.
+export function wrapArrayReplaceMarkersInStrings(body: string, mappings: KeywordMappings): string {
```
willvedd, line 53: This function does the heavy lifting in this PR. It solves the problem outlined in Problem 1 of the description.

test/context/yaml/context.test.js
```diff
@@ -541,23 +541,18 @@ describe('#YAML context validation', () => {
...
+      friendly_name: "##ENV## Tenant"
+      enabled_locales: @@LANGUAGES@@
+    connections:
+      - name: connection-1
+        strategy: waad
+        options:
+          tenant_domain: "##DOMAIN##"
+    `;
+
+    fs.writeFileSync(tenantFile, localAssets);
```
willvedd, line 555: Not possible to test this scenario with JSON, needs to be converted to YAML. Note the unwrapped `@@LANGUAGES@@` array replace instance.

sergiught (approved):

PR #766, opened 2023-03-14
Add CLI options with explanations

ianchanning:
### 🔧 Changes

Add CLI options with explanations.

Also switched `-c` to `--config_file` to give the long form of the options in the example `a0deploy export`.

Mostly the description is copied from `args.ts` with some additional explanation

### 📝 Checklist

- [ x] (N/A) All new/changed/fixed functionality is covered by tests (or N/A)
- [ x] I have added documentation for all new/changed functionality (or N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [7.17.2] - 2023-04-19
+
+### Fixed
+
+- API error when no tenant flags defined [#780]
+- Keyword preservation in a resource's identifier fields [#784]
+
+## [7.17.1] - 2023-03-31
+
+### Fixed
+
+- Tenant-agnostic filenames for client grants if more than fifty clients and fifty resource servers [#764]
+- Unintentional exclusion of clients when injecting access token via `AUTH0_ACCESS_TOKEN` [#775]
+
 ## [7.17.0] - 2023-03-03
 
 ### Added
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -76,7 +76,7 @@ In order for the Deploy CLI to call the Management API, a dedicated Auth0 applic
    c. Click “Authorize”
 
 > **Warning**
-> The Deploy CLI's own client grant is unconfigurable by itself to [prevent potentially destructive changes](./docs/resource-specific-documentation.md#client-grants).
+> The Deploy CLI's own client is unconfigurable by itself through the CLI to [prevent potentially destructive changes](./docs/resource-specific-documentation.md#client-grants).
```
willvedd, line 79: This tool has two modes of operation: CLI and node module. Don't want to cause any confusion here.

```suggestion
> The Deploy CLI's own client is unconfigurable by itself to [prevent potentially destructive changes](./docs/resource-specific-documentation.md#client-grants).
```

README.md
```diff
@@ -96,18 +96,44 @@ These values can be found in the “Settings” tab within the Auth0 application
...
 ```shell
 a0deploy export --format=yaml --output_folder=local
 ```
 
-Once the process completes, observe the resource configuration files generated in the `local` directory. Then, run the import command, which pushes configuration from the local machine to your Auth0 tenant:
+Once the process completes, observe the resource configuration files generated in the `local` directory. Then, run the `import` command, which pushes configuration from the local machine to your Auth0 tenant, here using a `config_file` instead of environment variables:
 
 ```shell
-a0deploy import -c=config.json --input_file local/tenant.yaml
+a0deploy import --config_file=config.json --input_file local/tenant.yaml --env=false
```
willvedd, line 108: The behavior of environment variables is made clear in this section: https://github.com/auth0/auth0-deploy-cli/blob/master/docs/configuring-the-deploy-cli.md#environment-variables. 

Because the README is meant to be more of a starting point, I'd rather not introduce concepts outside whats necessary to getting started.

willvedd:
Thanks for taking the time to contribute. Perhaps it needs to be made more clear, but the README is just a starting point, there are more detailed docs in the `/docs` folder. The content you propose appears to be redundant to that within the docs. If you believe that your changes still have standalone value, I would consider reviewing those pages and revising accordingly.

README.md
ianchanning: This could be put back - I was under the impression that if you used the `config_file` then you would also want to set `--env=false`.

ianchanning:
@willvedd thanks!

Issue #769, opened 2023-03-17
Issues using package as a node module 


NotMyself:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I am using the deploy cli as a node module in a visual studio code extension. I recently updated to the latest version and it appears that something has changed that prevents me from successfully loading a yml file to deploy. The code is relatively simple:

```typescript
try {
      const opts = {
        input_file:  filePath,
        config: await this.mergeConfig(fileDir, {
          AUTH0_ACCESS_TOKEN: accessToken,
          AUTH0_DOMAIN: getDomainFromToken(accessToken),
          AUTH0_ALLOW_DELETE: false,
          AUTH0_DEBUG: true,
        }),
        env: process.env,
      };
      await deploy(opts);
    } catch (e: any) {
      vscode.window.showErrorMessage(e.message);
    }
```

With this example the output gives no indication the file was not successfully loaded and suggests that I need to set the `AUTH0_ALLOW_DELETE` flag to `true`.

I have tried passing the base path both as an option named `base_path` and as a config setting named `AUTH0_BASE_PATH` with the `input_file` option set as a relative path tot he base path.

I suspect the module is using `cwd` somewhere deep in the code base and overriding what I am setting with the current working directory of VSCode not the project loaded in VSCode.

### Expectation

I expect the yml file to be loaded correctly or a message to be displayed that the file was not found that includes the location it attempted to load it from.

```
2023-03-17T17:04:23.983Z - info: Getting access token for undefined/23-03-07-vscode.us.auth0.com

2023-03-17T17:04:25.842Z - warn: Detected the following clients should be deleted. Doing so may be destructive.
[...]

PR #775, opened 2023-03-28
Fix: Management of clients when injecting access token

willvedd:
### 🔧 Changes

As reported by #769, there were issues managing clients when injecting an access token directly into the configuration as opposed to client credentials. This is because the exclusion logic was incorrectly filtering-out clients from being created, updated and deleted if the `AUTH0_CLIENT_ID` was not defined.

The simplest fix here is to not allow `currentClient` to be undefined by setting it to empty string if `AUTH0_CLIENT_ID` is not set. That way, when the exclusion comparison is executed, it will not evaluate to true if both it and `client_id` are undefined. 

### 📚 References

Original GH issue: #769 

### 🔬 Testing

Added new unit test.

### 📝 Checklist

### DIFF
--- a/src/tools/deploy.ts
+++ b/src/tools/deploy.ts
@@ -11,7 +11,11 @@ export default async function deploy(
   // Setup log level
   log.level = process.env.AUTH0_DEBUG === 'true' ? 'debug' : 'info';
 
-  log.info('Getting access token for ' + config('AUTH0_CLIENT_ID') + '/' + config('AUTH0_DOMAIN'));
+  log.info(
+    `Getting access token for ${
+      config('AUTH0_CLIENT_ID') !== undefined ? `${config('AUTH0_CLIENT_ID')}/` : ''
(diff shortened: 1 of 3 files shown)


src/tools/deploy.ts
```diff
@@ -11,7 +11,11 @@ export default async function deploy(
   // Setup log level
   log.level = process.env.AUTH0_DEBUG === 'true' ? 'debug' : 'info';
 
-  log.info('Getting access token for ' + config('AUTH0_CLIENT_ID') + '/' + config('AUTH0_DOMAIN'));
+  log.info(
+    `Getting access token for ${
+      config('AUTH0_CLIENT_ID') !== undefined ? `${config('AUTH0_CLIENT_ID')}/` : ''
```
willvedd, line 16: While I was at it, I fixed the logging statement here to avoid it printing out "undefined" if `AUTH0_CLIENT_ID` is not explicitly defined.

Issue #781, opened 2023-04-17
Crash when deploying in Bitbucket CI/CD environment with more than 3 clients


yossarian21:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

We have an Auth0 environment with about 8 Clients and are using the auth0-deploy-cli to provision multiple Auth0 tenants. I have been migrating the deployment script from "run it on a laptop" to proper CI/CD in Bitbucket Pipelines.

I have a simple Node script that retrieves our configuration (e.g. client id/client secret) and then invokes the CLI via a module:

```typescript
    await deploy({
        input_file: path.resolve(`${__dirname}/../config/tenant.yaml`),
        config,
    });
```

This has been working fine when running locally in a Docker containerized environment; however, in Bitbucket Pipelines, I was receiving the following crash. If I comment out most of our tenants, I can run the script successfully in Pipelines. Once I have configured more than 3 clients, I start getting the crash:

```shell
2023-04-14T19:38:49.084Z - info: Getting access token for <clientid>/<tenant>.us.auth0.com
2023-04-14T19:38:53.873Z - info: Updated [rules]: {"name":"auth0-account-link-extension","order":1}
2023-04-14T19:38:54.345Z - info: Updated [resourceServers]: {"name":"...","identifier":"..."}
2023-04-14T19:38:54.586Z - info: Updated [clients]: {"name":"Auth0 Management Client","client_id":"..."}
/usr/local/bin/node[290]: ../src/async_wrap.cc:324:static void node::AsyncWrap::QueueDestroyAsyncId(const v8::FunctionCallbackInfo<v8::Value>&): Assertion `args[0]->IsNumber()' failed.
[...]

PR #782, opened 2023-04-17
Add ability to override API concurrency (Addresses #781)

yossarian21:
### 🔧 Changes

In Bitbucket Pipelines CI/CD environment, API concurrency of higher than 1 causes an unexplained Node process crash (see #781). This PR allows me to override the API concurrency via an environment variable `AUTH0_API_CONCURRENCY`. If I run a Pipeline with a value of 1, it succeeds.

### 📚 References

https://github.com/auth0/auth0-deploy-cli/issues/781

### 🔬 Testing

Create tenant config containing more than 3 clients and attempt to deploy from a Bitbucket Pipeline. If you set `AUTH0_API_CONCURRENCY=1`, it will succeed.
We have also seen some crashes on Macs, but this is less reproducible.

### 📝 Checklist

- [N/A] All new/changed/fixed functionality is covered by tests (or N/A)
- [?] I have added documentation for all new/changed functionality (or N/A)

### DIFF
--- a/src/tools/auth0/client.ts
+++ b/src/tools/auth0/client.ts
@@ -11,7 +11,7 @@ import {
   BaseAuth0APIClient,
 } from '../../types';
 
-const API_CONCURRENCY = 3;
+const API_CONCURRENCY = process.env.AUTH0_API_CONCURRENCY ? parseInt(process.env.AUTH0_API_CONCURRENCY, 10) : 3;
 // To ensure a complete deployment, limit the API requests generated to be 80% of the capacity
 // https://auth0.com/docs/policies/rate-limits#management-api-v2
 const API_FREQUENCY_PER_SECOND = 8;


willvedd:
I'm glad that you've found success with this fork but I'd prefer to investigate a more appropriate specific to your Bitbucket integration than crowd and already crowded public API. Let's see if we can investigate more, I'd like to find a better solution for you and your team.

willvedd:
Thanks for the PR, I'm glad that folks feel empowered to contribute!

However, I don't agree with the technical advantage here. It appears that the proposed configuration is a workaround for a personal CI limitation, not necessarily one that would be of benefit to other users. The Deploy CLI already has numerous configurations and I'm hesitant to add bloat. Further, the default value of 3 concurrent requests is a sensible default that strikes a balance between performance and reliability.

Issue #770, opened 2023-03-20
Keyword preservation does not preserve "identifier" fields


dkobia:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

The new keyword replacement works as required (mostly) but skips a few parameters.

For example, the following will not be preserved in tenant.yml

```yaml
clientGrants:
  - audience: '@@AUTH0_AUDIENCE@@'
```

```yaml
themes:
  - fonts:
      font_url: '@@FONT1_URL@@'
```

```yaml
themes:
  - widget:
      logo_url: '@@LOGO_URL@@'
```

The same is true for any keyword mappings in `branding_templates/universal_login.html`

### Expectation

I expect the `AUTH0_KEYWORD_REPLACEMENT_MAPPINGS` to be maintained.

### Reproduction

1. Export tenant settings that include changes to the universal login template
2. Update keyword mappings in the html file
3. Run the import again

### Deploy CLI version

v7.17.0

### Node version

v16.10.0

willvedd:
@dkobia thank you very much for the feedback! I'm glad that this feature is getting a bit of use. I believe that there are two separate issues at play here.

First is that there probably are a couple of fields that the algorithm is not yet able to preserve, especially when they are fields nested underneath arrays (ex: themes). This is because array items cannot be identified by order alone (for technical reasons explained in #688). This is only my initial suspicion and I'll need to investigate further to definitively identify the specific issue.

[...]

PR #784, opened 2023-04-18
DXCDT-421: Preserve identifier fields

willvedd:
### 🔧 Changes

As noted by #770 , there are cases where users want to preserve “identifier fields” the recently added keyword preservation functionality. This was not previously implemented because of the complexity paired with the perceived unlikeliness of this needing to happen. However, it is clear that this functionality is desired by customers.

This PR requires a few modifications to the keyword preservation algorithm but mostly test cases have been added or updated.

**Example:** 
For clients, the `name` field is the defacto identifier across environments.
```yaml
clients:
  - name: "Auth0 CLI - ##ENV##"
    allowed_clients: []
    app_type: non_interactive
    callbacks: []
    client_aliases: []
    cross_origin_auth: false
```

A limitation to note is that this will not work if the keyword value on remote is different than that of the local value. For example, if you have a `##ENV##` keyword with the local value of `dev` and the dump is occurring on production (`prod`), the keyword preservation of identifier fields **will not succeed**.

### 📚 References

Original Thread: #770 

### 🔬 Testing

Several new unit and integration test cases.

### 📝 Checklist

### DIFF
--- a/src/keywordPreservation.ts
+++ b/src/keywordPreservation.ts
@@ -157,10 +157,21 @@ export const convertAddressToDotNotation = (
 
 export const updateAssetsByAddress = (
   assets: object,
-  address: string,
+  addresses: [string, string],
   newValue: string
 ): object => {
-  const dotNotationAddress = convertAddressToDotNotation(assets, address);
+  const dotNotationAddress = (() => {
+    //Two address possibilities are provided to account for cases when there is a keyword
+    //in both the resources's identifier field and another nested field. This is because
+    //when the resource identifier's field is preserved on the remote assets tree, it loses
+    //its identify, so we'll need to try two addresses: one where the identifier field has
+    //a keyword and one where the identifier field has the literal replaced value
+    const possibility1 = convertAddressToDotNotation(assets, addresses[0]);
+    const possibility2 = convertAddressToDotNotation(assets, addresses[1]);
+
+    if (possibility1 !== null) return possibility1;
+    return possibility2;
+  })();
 
   if (dotNotationAddress === null) return assets;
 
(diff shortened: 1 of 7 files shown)


src/keywordPreservation.ts
```diff
@@ -91,7 +91,7 @@ export const getAssetsValueByAddress = (address: string, assets: any): any => {
 
   // It is easier to handle an address piece-by-piece by
   // splitting on the period into separate "directions"
-  const directions = address.split('.');
+  const directions = address.split(/\.(?![^\[]*\])/g);
```
willvedd, line 94: Splits periods unless they're within square brackets `[]`. This is because with this new functionality, the address may have a URL or value with a period in it. Example: `customDomains.[domain=travel0.com].domain`

src/keywordPreservation.ts
```diff
@@ -213,7 +213,15 @@ export const preserveKeywords = ({
...
   addresses.forEach((address) => {
     const localValue = getAssetsValueByAddress(address, localAssets);
-    const remoteValue = getAssetsValueByAddress(address, remoteAssets);
+
+    const remoteAssetsAddress = (() => {
+      const doesAddressHaveKeyword = doesHaveKeywordMarker(address, keywordMappings);
+      if (doesAddressHaveKeyword) {
+        return keywordReplace(address, keywordMappings);
+      }
+      return address;
```
willvedd, line 222: This conditional check isn't completely necessary but it reduces the blast radius of any issues that could occur if an erroneous keyword replacement happened. Also, it serves to document why two codepaths exist.

src/keywordPreservation.ts
```diff
@@ -10,7 +10,7 @@ import APIHandler from './tools/auth0/handlers/default';
   Original Github Issue: https://github.com/auth0/auth0-deploy-cli/issues/328
 */
 
-export const shouldFieldBePreserved = (
+export const doesHaveKeywordMarker = (
```
willvedd, line 13: Renamed to apply more generically across codebase.

sergiught (approved):

Issue #786, opened 2023-04-20
I have exactly the same issue as https://github.com/auth0/auth0-deploy-cli/issues/88 


RealJohnnyBravo:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I have exactly the same issue as https://github.com/auth0/auth0-deploy-cli/issues/88
Tried all the solutions mentioned in that ticket - nothing helped.
I need manually go to auth0 page and do some change like add and then remove space - only in such case auth0 does not show me "callbacks" must be a valid uri auth0.
M setup:
    "CALLBACKS": "http://localhost:3000/api/auth/callback/auth0, https://test1.com/api/auth/callback/auth0, https://test2.com/api/auth/callback/auth0, https://test3.com/api/auth/callback/auth0",

  "callbacks": [
    "@@CALLBACKS@@"
    ]

Also tried with hashes and with "CALLBASKS" as array.
```

### Expectation

No manual intervention is needed.

### Reproduction

Provided in the description

### Deploy CLI version

N/A

### Node version

N/A

**Comment by willvedd** [comment]
This works just fine for me:

```json
{
  "callbacks": "@@ALLOWED_CALLBACKS@@"
}
```

RealJohnnyBravo:
If I do as suggested, I am getting this error: 

Bad Request: Payload validation error: 'Expected type array but found type string' on property allowed_logout_urls (URLs that are valid to redirect to after logout from Auth0). (also) Payload validation error: 'Expected type array but found type string' on property callbacks (A set of URLs that are valid to call back from Auth0 when authenticating users).

PR #787, opened 2023-04-24
Adding directory example to keyword replacement docs

willvedd:
### 🔧 Changes

As made aware by #786 , there are no examples of keyword replacement for the directory format, only YAML. While the implementations are very similar, there are notable differences so providing an example should help clear those up.

### 📚 References

Original issue: #786

### DIFF
--- a/docs/keyword-replacement.md
+++ b/docs/keyword-replacement.md
@@ -16,7 +16,7 @@ To use the keyword replacement, the `AUTH0_KEYWORD_REPLACEMENT_MAPPINGS` configu
   "AUTH0_CLIENT_SECRET": "BAR",
   "AUTH0_KEYWORD_REPLACE_MAPPINGS": {
     "ENVIRONMENT": "dev",
-    "ALLOWED_LOGOUTS": ["https://dev-test-site.com/logout", "localhost:3000/logout"],
+    "ALLOWED_LOGOUT_URLS": ["https://dev-test-site.com/logout", "localhost:3000/logout"],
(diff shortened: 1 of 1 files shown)


docs/keyword-replacement.md
```diff
@@ -26,14 +26,23 @@ To use the keyword replacement, the `AUTH0_KEYWORD_REPLACEMENT_MAPPINGS` configu
 
 ```yaml
 tenant:
-  friendly_name: My ##ENVIRONMENT## tenant
-  allowed_logout_urls: @@ALLOWED_LOGOUTS@@
+  friendly_name: "##ENVIRONMENT## tenant"
```
willvedd, line 29: Adding an example with wrapped quotes shows how to implement string replace keyword when there is no preceding string to the keyword.

docs/keyword-replacement.md
```diff
@@ -16,7 +16,7 @@ To use the keyword replacement, the `AUTH0_KEYWORD_REPLACEMENT_MAPPINGS` configu
   "AUTH0_CLIENT_SECRET": "BAR",
   "AUTH0_KEYWORD_REPLACE_MAPPINGS": {
     "ENVIRONMENT": "dev",
-    "ALLOWED_LOGOUTS": ["https://dev-test-site.com/logout", "localhost:3000/logout"],
+    "ALLOWED_LOGOUT_URLS": ["https://dev-test-site.com/logout", "localhost:3000/logout"],
```
willvedd, line 19: Slight renaming of keyword because "logouts" is not a technically correct term.

sergiught (approved):

PR #792, opened 2023-05-18
Update tenant migration flag list

mtj8033:
### 🔧 Changes
Was receiving the following exception when promoting our code through envs in some of our tenants:

Updating the enforce_client_authentication_on_passwordless_start flag is not allowed. Please contact us if you need additional information.

Added in that plus any other ones listed on the management api docs. Let me know if this should just do the one I actually hit it on versus all of them.

### 📚 References
Older PR that added this capability - https://github.com/auth0/auth0-deploy-cli/pull/590

### 🔬 Testing
Verify on tenants that don't have the added flags that they are stripped out from the list. I didn't update any unit test as they already covered this general use case, but let me know if I should have updated something.

### 📝 Checklist

- [X] All new/changed/fixed functionality is covered by tests (or N/A) - See above comment from testing
- [X] I have added documentation for all new/changed functionality (or N/A) - N/A?


willvedd:
Thanks for this contribution! As you've seen, it's pretty straightforward but I want to take a moment to vet these settings. Also, the HTTP mocks in our tests to be "re-recorded" to reflect the payload changes, but that's a simple thing for me to do. Please allow me a few days to get to this. But again, thanks for the PR.

DropnFbombs:
Can the value "new_universal_login_experience_enabled" please be added to this merge.  

I discovered today that the CLI is throwing back this same error but for that value.  It is not listed in the Auth0 Management API documentation however it does work.  

I don't know if by it not being listed in the documentation I should stop using it or if it's an oversight on their end, however I can confirm that the value is included in the tenant.yaml file when I pull a tenant and that it works with the API when I push that value using api/v2/tenants/settings

willvedd:
@mtj8033 Thanks for the contribution, I'm happy that you felt empowered to submit! We've had a recent increase in the number of issues around tenant flags and have had to rethink how we handle these. The solution that we've come up with has changed the fundamental mechanism for how flags are sanitized from one tenant to the next (see #796). This means that this PR is no longer relevant, so I will be closing. But thanks again for aiding in the discussion.

PR #793, opened 2023-05-19
Fix: Preserving client grants' `audience` field

willvedd:
### 🔧 Changes

Internal issue raised (ESD-28198) that reported that `clientGrants[].audience` fields weren't being preserved on export even after #784 was released. This bug was occurring because a client grant does not have a single identifier field but two that create a "compound" identifier field that ensures uniqueness across tenants; client grants are the only resource that have this property. This logic was not accounted for in previous iterations of keyword preservation but is added in this PR.

### 📚 References

Previous PR that added identifier field preservation: #784 

### 🔬 Testing

Unit tests added to the primary `preserveKeywords` function tests but to also to the underlying `getPreservableFieldsFromAssets` function too.

### 📝 Checklist

### DIFF
--- a/src/keywordPreservation.ts
+++ b/src/keywordPreservation.ts
@@ -208,17 +208,15 @@ export const preserveKeywords = ({
 }): object => {
   if (Object.keys(keywordMappings).length === 0) return remoteAssets;
 
-  const resourceSpecificIdentifiers = auth0Handlers.reduce(
-    (acc, handler): Partial<{ [key in AssetTypes]: string[] }> => {
+  const resourceSpecificIdentifiers: Partial<{ [key in AssetTypes]: string[] }> =
+    auth0Handlers.reduce((acc, handler): Partial<{ [key in AssetTypes]: string[] }> => {
       return {
         ...acc,
         [handler.type]: handler.identifiers.filter((identifiers) => {
           return identifiers !== handler.id;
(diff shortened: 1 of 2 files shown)


src/keywordPreservation.ts
```diff
@@ -186,21 +202,21 @@ export const preserveKeywords = ({
   localAssets: object;
   remoteAssets: object;
   keywordMappings: KeywordMappings;
-  auth0Handlers: Pick<APIHandler, 'id' | 'identifiers' | 'type'>[];
+  auth0Handlers: (Pick<APIHandler, 'id' | 'type'> & {
+    identifiers: (string | string[])[];
+  })[];
```
willvedd, line 207: There is a morass of logic elsewhere in the codebase that made it untenable to modify the type directly on the `APIHandler`'s `identifiers` field.

test/keywordPreservation.test.ts
```diff
@@ -658,12 +696,6 @@ describe('preserveKeywords', () => {
       ],
     });
 
-    const expected = (() => {
-      let expected = mockLocalAssets;
-      expected.actions = [mockLocalAssets.actions[0]];
-      return expected;
-    })();
-
```
willvedd, line 666: Unused code.

sergiught (approved):

Issue #788, opened 2023-04-27
Export from a new tenant includes non-permitted flag require_signed_request_object


yossarian21:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I set up a new tenant and exported config from it using `format: 'yaml'`.

The resulting `tenant.yaml` includes:

```yaml
tenant:
  flags:
    require_signed_request_object: false
```

If I immediately try to import this into the same tenant, I get an error:

```shell
Bad Request: Updating the require_signed_request_object flag is not allowed. Please contact us if you need additional information.
    at fn (/app/node_modules/rest-facade/src/Client.js:402:25)
    at Request.callback (/app/node_modules/superagent/src/node/index.js:923:3)
    at fn (/app/node_modules/superagent/src/node/index.js:1153:18)
    at IncomingMessage.<anonymous> (/app/node_modules/superagent/src/node/parsers/json.js:19:7)
    at IncomingMessage.emit (node:events:525:35)
    at IncomingMessage.emit (node:domain:489:12)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at processTicksAndRejections (node:internal/process/task_queues:82:21) {
  statusCode: 400,
  requestInfo: {
    method: 'patch',
    url: 'https://<tenant>.us.auth0.com/api/v2/tenants/settings'
  },
```

I can't find this flag in the tenant settings UI (or I don't understand what it is) but I haven't modified any of the tenant-level settings other than inviting members, setting a Friendly Name, and setting the Logo URL.

Workaround: Remove `require_signed_request_object` from `flags`

### Expectation

I should be able to re-import the config that was just exported from a new tenant.

### Reproduction

[...]

PR #796, opened 2023-06-05
Update list of tenant migration flags

willvedd:
### 🔧 Changes

There has been a recent uptick in incompatible tenant flags resulting in errors that resemble the following ([source](https://community.auth0.com/t/additional-properties-not-allowed-new-universal-login-experience-enabled/108069)):

```
Payload validation error: 'Additional properties not allowed: new_universal_login_experience_enabled' on property flags (Flags used to change the behavior of this tenant).
```

The above example specifically calls out the `new_universal_login_experience_enabled` flag but this issue could occur for a handful of other properties too. 

The issue here is that deprecations occur over time and sometimes these deprecations have a lag of rollout between environments and regions. What this means is that it is very possible for customers to have old config that successfully applies for one tenant but does not apply for another. The vast majority of the time, customers do not need to concern themselves with these tenant flags, they primarily exist to facilitate work behind the scenes.

This PR adds a few of the known problematic tenant flags that have been reported. This solution is more of a **temporary patch** to smooth over customer issues while a broader investigation around tenant flag management can be carried out. As a future note to folks, if anyone is continuing to receive this error after the release of this PR it is safe to remove the error-causing property from your `tenant.yaml` or `tenant.json` configuration.

### 📚 References

- [Related Community Thread](https://community.auth0.com/t/additional-properties-not-allowed-new-universal-login-experience-enabled/108069)
- [Related Github issue](https://github.com/auth0/auth0-deploy-cli/issues/788)

### 🔬 Testing

Added a few more test cases to prove that this will work.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/tenant.ts
+++ b/src/tools/auth0/handlers/tenant.ts
@@ -105,6 +105,9 @@ export const sanitizeMigrationFlags = ({
     'trust_azure_adfs_email_verified_connection_property',
     'include_email_in_reset_pwd_redirect',
     'include_email_in_verify_email_redirect',
+    'new_universal_login_experience_enabled',
+    'require_signed_request_object',
+    'enforce_client_authentication_on_passwordless_start',
   ];
 
   return Object.keys(proposedFlags).reduce(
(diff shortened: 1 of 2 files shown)


sergiught (approved):

willvedd:
Closing this out in favor of #797

Issue #606, opened 2022-07-21
Exeeding Management API rate limit


TAKANET7007:
### Description
Dump metod makes the use of api exceeds the limit of Management API endpoint.
The dump operation itself seems to worked fine, but there was a log left on Auth0 console saying "Rate Limit On API"
This was not happening at v7.11.1.

### Reproduction
Our code looks almost exactly like one in the document.
https://github.com/auth0/auth0-deploy-cli/blob/master/docs/using-as-node-module.md

Following is the log shown at auth0 console.(I excluded some records)
```
{
    "type":"api_limit",
    "description": "Global per socond default group limit has been reached",
    "details":{
        "request": {
            "method": "get",
            "path": "/api/v2/prompts/{prompt}/custom-text/{language}"
        },
       "response": {}
    }
}
```

From the log above, this seems to be related to custom prompts.
I have only one language activated but some of the prompts are customesed on my tenant.

### Environment
- **Version of this library used:** 7.14.1
- **Version of the platform or framework used, if applicable:** Node.js 16.x
- **Other relevant versions (language, server software, OS, browser):** AWS lambda
- **Other modules/plugins/libraries that might be involved:**

willvedd:
It is normal and expected for the Deploy CLI to be operating very near or at the rate-limiting threshold. Doing so allows the tool to perform as best as it can by relying on the underlying SDK's retry logic to appropriately manage the flow of requests to the management API. This is especially true in more recent versions that support custom text prompts, as that resource type necessarily requires many dozens of requests to comprehensively build up your translations.

TAKANET7007:
Hi

Thank you for the reply.
I agree that the retry logic will make things go right for this CLI.
But it is true that few end user will experience short service outage when rate limit occur if there is a fucntion dependent on management API.
It seems you have 'promise-pool-executor' for job management.
Is it possible not to make API go off the limit by changing API_FREQUENCY_PER_SECOND?
I tried to read through the code but I was not able to find how the requests for the "prompts" are made and managed.

PR #804, opened 2023-06-22
DXCDT-476: Connection pool manager integration for custom text prompts API requests

willvedd:
### 🔧 Changes

The Deploy CLI is greedy with respect to custom text prompts API requests. In the case of tenants with several languages, it conceivably can create hundreds or even over one thousand API requests to fetch and update that configuration.

While we would simply expect the API to rate-limit the application in times of high volume, sometimes that sudden high-volume surfaces some nefarious issues. In practice this is evident through socket time-out and SSL handshake termination errors; not application-layer errors. To help reduce the occurrence of these load-causing issues, the Deploy CLI should leverage the existing connection pool handler for batching and generally facilitating a more controlled execution of these requests.

### 📚 References

Related Github issue: #606 

### 🔬 Testing

Passes end-to-end tests.

**NOTE:** Some unit tests have been skipped. This is because the connection pool handler is difficult to mock.

### 📝 Checklist


src/types.ts
```diff
@@ -187,14 +188,7 @@ export type BaseAuth0APIClient = {
...
 export type Auth0APIClient = BaseAuth0APIClient & {
-  pool: {
-    addSingleTask: (arg0: { data: Object; generator: any }) => {
-      promise: () => Promise<ApiResponse>;
-    };
-    addEachTask: (arg0: { data: Object; generator: any }) => {
-      promise: () => Promise<Asset[][]>;
-    };
-  };
+  pool: PromisePoolExecutor;
```
willvedd, line 191: Adds more robust typing to the connection pool's `pool` property. All type modifications elsewhere in this PR are related to this change.

test/tools/auth0/handlers/prompts.tests.ts
```diff
@@ -10,7 +10,7 @@ const mockPromptsSettings = {
 
 describe('#prompts handler', () => {
   describe('#prompts process', () => {
-    it('should get prompts settings and prompts custom text', async () => {
+    it.skip('should get prompts settings and prompts custom text', async () => {
```
willvedd, line 13: Mocking the connection pool manager is tricky here and my initial attempts were unsuccessful. Hence why this PR is in draft.

sergiught (approved):

PR #887, opened 2024-02-02
Upgrading node-auth0 from v3 to v4

willvedd:
### 🔧 Changes

Upgrading `node-auth0` from v3 to v4. This is a necessary upgrade because all new Auth0 features and dependency patches are getting released on v4 now.

This was originally attempted in #852 but had a few tricky changes that need to be reviewed. This should _not_ require a major version bump for Deploy CLI despite the breaking changes in v4. There will however be a few awkward situations that will need to be smoothed-over. 

### 📚 References

- Original PR: #852 

### 📝 Checklist


src/context/directory/handlers/clientGrants.ts
```diff
@@ -47,13 +47,14 @@ async function dump(context: DirectoryContext): Promise<void> {
 
   if (clientGrants.length === 0) return;
 
-  const allResourceServers = await context.mgmtClient.resourceServers.getAll({
-    paginate: true,
+  // TODO: Bring back paginate: true
+  const { data: { resource_servers: allResourceServers } } = await context.mgmtClient.resourceServers.getAll({
```
willvedd, line 51: Yes, this is important to include. Before Node v4, Auth0 Node SDK didn't natively support pagination and I don't believe that v4 added it either but this needs to be confirmed. The `getAll` function is mutated by the Deploy CLI to support the `paginate` property.

src/context/index.ts
```diff
@@ -161,19 +161,19 @@ export const setupContext = async (
...
     })();
 
-    const clientCredentials = await authClient.clientCredentialsGrant({
+    const clientCredentials = await authClient.oauth.clientCredentialsGrant({
       audience: config.AUTH0_AUDIENCE
         ? config.AUTH0_AUDIENCE
         : `https://${config.AUTH0_DOMAIN}/api/v2/`,
     });
-    return clientCredentials.access_token;
+    return clientCredentials.data?.access_token;
```
willvedd, line 176: Should introduce some checks to prevent an undefined `access_token` from being persisted through the rest of the process.

src/tools/auth0/client.ts
```diff
@@ -167,7 +167,7 @@ export default function pagedClient(client: BaseAuth0APIClient): Auth0APIClient
       frequencyLimit: API_FREQUENCY_PER_SECOND,
       frequencyWindow: 1000, // 1 sec
     }),
-  };
+  } as unknown as Auth0APIClient;
```
willvedd, line 170: Looks very weird. Can the `unknown` be removed or this type cast otherwise cleaned up?

src/tools/auth0/handlers/actions.ts
```diff
@@ -92,29 +93,34 @@ export function isMarketplaceAction(action: Action): boolean {
...
+  async createAction(action: PostActionRequest) {
     // Strip the deployed flag
     const addAction = { ...action };
-    delete addAction.deployed;
-    delete addAction.status;
-    const createdAction = await this.client.actions.create(addAction);
 
+    // TODO: Should we keep this?
+    delete (addAction as any).deployed;
+    delete (addAction as any).status;
```
willvedd, line 116: Yes because when you create an action you cannot specify the `deploy` or `status` in the payload. However, this should be accomplished with the `stripCreateFields: ['deployed', 'status'],` in the handler definition above.

src/tools/auth0/handlers/actions.ts
```diff
@@ -202,7 +208,9 @@ export default class ActionHandler extends DefaultAPIHandler {
     // Actions API does not support include_totals param like the other paginate API's.
     // So we set it to false otherwise it will fail with "Additional properties not allowed: include_totals"
     try {
-      const actions = await this.client.actions.getAll({ paginate: true });
+      // TODO: bring back paginate: true
+      const { data } = await this.client.actions.getAll();
```
willvedd, line 212:
```suggestion
      const { data: {actions} } = await this.client.actions.getAll();
```

src/tools/auth0/handlers/emailProvider.ts
```diff
@@ -34,9 +36,10 @@ export default class EmailProviderHandler extends DefaultHandler {
 
     let existing = await this.getType();
 
+    // TODO: HTTP DELETE on emails/provider is not public, so not part of our vNext SDK.
     if (Object.keys(emailProvider).length === 0) {
       if (this.config('AUTH0_ALLOW_DELETE') === true) {
-        await this.client.emailProvider.delete();
+        // await this.client.emails.delete();
```
willvedd, line 42: This could technically be a breaking change but it also seems like this never worked to begin with. We may be able to replicate a delete-like functionality to the best of our ability.

src/tools/auth0/handlers/emailProvider.ts
```diff
@@ -45,20 +48,18 @@ export default class EmailProviderHandler extends DefaultHandler {
...
     if (existing.name) {
-      const provider = { name: emailProvider.name, enabled: emailProvider.enabled };
-      const updated = await this.client.emailProvider.update(provider, emailProvider);
+      const updated = await this.client.emails.update(emailProvider);
       this.updated += 1;
       this.didUpdate(updated);
     } else {
-      const provider = { name: emailProvider.name, enabled: emailProvider.enabled };
-      const created = await this.client.emailProvider.configure(provider, emailProvider);
+      const created = await this.client.emails.configure(emailProvider as PostProviderRequest);
```
willvedd, line 62: Can this type be applied to the construction of the `emailProvider` value above rather than smoothing-over as a type cast?

src/tools/auth0/handlers/emailTemplates.ts
```diff
@@ -51,16 +51,18 @@ export default class EmailTemplateHandler extends DefaultHandler {
     try {
       const identifierField = this.identifiers[0];
 
-      const params = { name: emailTemplate[identifierField] };
-      const updated = await this.client.emailTemplates.update(params, emailTemplate);
-      delete updated.body;
+      const params = { templateName: emailTemplate[identifierField] };
+      const { data: updated } = await this.client.emailTemplates.update(params, emailTemplate);
+      // TODO: Should we revisit this?
+      delete (updated as any ).body;
```
willvedd, line 57: @frederikprijck What is there to revisit?

src/tools/auth0/handlers/emailTemplates.ts
```diff
@@ -51,16 +51,18 @@ export default class EmailTemplateHandler extends DefaultHandler {
...
       this.didUpdate(updated);
       this.updated += 1;
     } catch (err) {
       if (err.statusCode === 404) {
         // Create if it does not exist
-        const created = await this.client.emailTemplates.create(emailTemplate);
-        delete created.body;
+        const { data: created } = await this.client.emailTemplates.create(emailTemplate);
+        // TODO: Should we revisit this?
+        delete (created as any).body;
```
willvedd, line 65: Similarly, @frederikprijck what is there to revisit?

src/tools/auth0/handlers/guardianFactorProviders.ts
```diff
@@ -40,8 +40,19 @@ export default class GuardianFactorProvidersHandler extends DefaultHandler {
...
+        // TODO: This is quite a change, needs to be validated for sure.
+        if (m.name === 'phone' && m.provider === 'twilio') {
+          provider = await this.client.guardian.getPhoneFactorProviderTwilio();
+        } else if (m.name === 'sms' && m.provider === 'twilio') {
+            provider = await this.client.guardian.getSmsFactorProviderTwilio();
+        } else if (m.name === 'push-notification' && m.provider === 'apns') {
+            provider = await this.client.guardian.getPushNotificationProviderAPNS();
+        } else if (m.name === 'push-notification' && m.provider === 'sns') {
+          provider = await this.client.guardian.getPushNotificationProviderSNS();
+        }
```
willvedd, line 53: Verify if there are any other providers

src/types.ts
```diff
@@ -34,158 +30,14 @@ export type PagePaginationParams = SharedPaginationParams & {
...
-  };
-  tenant: APIClientBaseFunctions & {
-    getSettings: () => Promise<Tenant>;
-    updateSettings: (arg0: Partial<Tenant>) => Promise<Tenant>;
-  };
-  triggers: APIClientBaseFunctions & {
-    getTriggerBindings: () => Promise<Asset>;
-  };
-}; // TODO: replace with a more accurate representation of the Auth0APIClient type
+export type BaseAuth0APIClient = ManagementClient;
```
willvedd, line 40: Should remove the `BaseAuth0APIClient` type altogether.

src/tools/auth0/client.ts
willvedd: **Update:** Looking at this with Frederik again, we should probably find a way to exclude the pool property or facilitate pagination some other way that can retain the `Auth0APIClient` type.

src/tools/auth0/handlers/emailProvider.ts
```diff
@@ -16,7 +17,8 @@ export default class EmailProviderHandler extends DefaultHandler {
 
   async getType(): Promise<Asset> {
     try {
-      return await this.client.emailProvider.get({ include_fields: true, fields: defaultFields });
+      const { data  } = await this.client.emails.get({ include_fields: true, fields: defaultFields.join(',') });
```
willvedd, line 20: This looks like a breaking change but the SDK types `fields` as a string so this doesn't precludes the need to type cast.

src/tools/auth0/handlers/index.ts
```diff
@@ -20,7 +20,6 @@ import * as guardianPhoneFactorMessageTypes from './guardianPhoneFactorMessageTy
 import * as roles from './roles';
 import * as branding from './branding';
 import * as prompts from './prompts';
-import * as migrations from './migrations';
```
willvedd, line 23: Migrations not part of the public spec. Need to understand if this was an intentional deprecation or if it is even necessary.

src/tools/auth0/handlers/tenant.ts
```diff
@@ -1,16 +1,16 @@
...
-import { Asset, Assets, Language } from '../../../types';
+import { Asset, Assets } from '../../../types';
 import log from '../../../logger';
 
 export const schema = {
   type: 'object',
 };
 
-export type Tenant = Asset & { enabled_locales?: Language[]; flags: { [key: string]: boolean } };
+export type Tenant = TenantSettings;
```
willvedd, line 13: Would rather we just get rid of the `Tenant` type altogether in favor of the `TenantSettings`.

test/e2e/recordings/should-deploy-without-throwing-an-error.json
```diff
@@ -2,7 +2,7 @@
     {
         "scope": "https://deploy-cli-dev.eu.auth0.com:443",
         "method": "GET",
-        "path": "/api/v2/rules?include_totals=true&page=0&per_page=100",
```
willvedd, line 5: These recordings should probably be reverted once the pagination functionality is added back in.

test/e2e/e2e.test.ts
```diff
@@ -7,6 +7,10 @@ import { load as yamlLoad } from 'js-yaml';
 import { setupRecording, testNameToWorkingDirectory } from './e2e-utils';
 import { dump, deploy } from '../../src';
 import { AssetTypes } from '../../src/types';
+import fetch from 'node-fetch';
+
+// Use node.http for recording library
+global.fetch = fetch;
 
 const shouldUseRecordings = process.env['AUTH0_HTTP_RECORDINGS'] === 'lockdown';
 const AUTH0_DOMAIN = shouldUseRecordings
```
willvedd, line 16: Basically everthing below this line is an innocuous testing change 🙌

willvedd (commented):
Outstanding items with this PR:

- DELETE email provider SDK function – Review the impact of this. Can this be smoothed-over? Should this be re-added to the SDK? 
- Migrations removal – Review the impact of this. Do we actually need or want migrations? However, it is a large enough breaking change where if this was removed from the Deploy CLI, you would need to create a new major version
- Pagination – Bring back and probably favor a method of implementation that doesn't monkey-patch the SDK. Also look into relocating the `pool` construct. 
- Types – Node SDK v4 introduces comprehensive types, which should take precedence over the types defined in this repo. Lots of weird type casting going on, should review and used canonical types from SDK wherever possible. 
- Node Compatibility  – The Node SDK v4 drops support for v16 w/ introduction of `fetch`. We've dropped node version support in the past without much noise but need to review the usage decide the gravity if this change and if it warrants a new major release or not.

There are some other minor things to review in this PR but otherwise the above are the major chunks of work. IMO, we should attempt to not introduce a new major to facilitate this new SDK because there are many improvements that can be made with a new major that should be provided to developers.

src/context/directory/handlers/clientGrants.ts
willvedd: **Update:** This could be an opportunity to improve the pagination. I think that with the v4 SDK bringing types that monkey-patching the client like this could be problematic but if you're able to get it work then great. Otherwise may need to break out into a separate function or some other abstraction.

src/context/index.ts
willvedd: **Recommendation:** Ensure ample amount of undefined checks where appropriate.

src/tools/auth0/handlers/emailProvider.ts
willvedd: What I mean by this is that the typing can probably be improved. I don't immediately see how but I'm sure it can to avoid this messy type casting.

src/context/directory/handlers/clientGrants.ts
willvedd: **Update 2:** To be clear, the existing code should/will work, the only considerations is to get the types to match up. Because the SDK wasn't typed in v3, we had to create our own `BaseAuth0APIClient` type and augment with the `pool` property. The proposed code in this PR should actually be removed and probably instead replace the `BaseAuth0APIClient` with the v4 SDK's native management client type. There may be additional type mismatches but that _should_ get this to a mostly-functioning state.

robosung:
Hi - it looks like there have not been any activity on this PR for a while. Are there plan to bump to v4 soon? Thank you.

Issue #924, opened 2024-07-17
Deploy broken in v7.23.0: SCIM support for connection management requires read:scim_config scope or update to excluded_props


bweisberger:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/tag/auth0-deploy-cli) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

We successfully deployed as recently as two days ago, but suddenly our deploys are breaking with the following error:
`error: SCIM request failed with statusCode 403 (insufficient_scope). Insufficient scope, expected any of: read:scim_config.`

Looking at the auth0-deploy-cli repo, I can see [this PR](https://github.com/auth0/auth0-deploy-cli/pull/921) to add SCIM configuration management was merged yesterday. 

We use the a0deploy cli tool to deploy configuration to Auth0 from IaC, including our connections. Our deploy script installs the package without pinning a version, so I'm assuming it is installing the latest version: 
```sh
npm install -g auth0-deploy-cli
```
We don't have this permission enabled for our Auth0 Deploy CLI client in Auth0, and we don't have the `connections.scim_configuration` property in our excluded props. I'm reasoning we shouldn't have to make a code change without a major version update.

Thanks for your help!

### Expectation

Installing the latest minor version of the auth0-deploy-cli package should not require code updates or cause deploys to fail.

### Reproduction

1. Given your Auth0 Deploy CLI client does not have the `read:scim_config` scope, and `connections.scim_configuration` is not an excluded prop
2. When you download the latest minor version (7.23.0) of the auth0-deploy-cli npm package and run a0deploy import to deploy IaC
3. The deploy will fail with an error that the `read:scim_config` scope is required.

### Deploy CLI version

7.23.0

### Node version

16.20.2

[...]

PR #925, opened 2024-07-18
 Adding fixes for permission errors (403)

nandan-bhat:
### 🔧 Changes

- This pull request introduces checks to handle permission errors that arise when processing connections without `scim_config` permissions.
- If these errors occur, affected connections are skipped, and a warning message is displayed.
- Additionally, this PR resolves issues related to the `too_many_requests` (429) error.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/scimHandler.ts
+++ b/src/tools/auth0/handlers/scimHandler.ts
@@ -14,8 +14,10 @@ interface scimRequestParams {
 
 interface scimBodyParams {
 	user_id_attribute: string;
-	mapping: { scim: string; auth0: string; }[];
+	mapping: { scim: string; auth0: string }[];
 }
+
+interface ScimScopes { read: boolean; create: boolean; update: boolean; delete: boolean }
 
 /**
  * The current version of this sdk use `node-auth0` v3. But `SCIM` features are not natively supported by v3.
@@ -27,7 +29,18 @@ export default class ScimHandler {
 	private tokenProvider: any;
 	private config: any;
 	private connectionsManager: any;
-  private currentConnectionId: string | undefined = undefined;
+	private updateQueue: any[] = [];
+	private isExecuting = false;
+	private scimScopes: ScimScopes = { read: true, create: true, update: true, delete: true };
+	private scopeMethodMap = {
+		get: 'read',
+		post: 'create',
+		patch: 'update',
+		delete: 'delete'
+	}
+
+	expectedChanges: number = 0;
+	completedChanges: number = 0;
     
 	constructor(config, tokenProvider, connectionsManager) {
 		this.config = config;
@@ -55,11 +68,12 @@ export default class ScimHandler {
 		this.idMap.clear();
 
 		for (let connection of connections) {
+			if (!this.scimScopes.read) return;
 			if (!this.isScimStrategy(connection.strategy)) continue;
       
       // To avoid rate limiter error, we making API requests with a small delay.
       // TODO: However, this logic needs to be re-worked.
-      await sleep(500);
+			await sleep(250);
 
       this.idMap.set(connection.id, { strategy: connection.strategy, hasConfig: false });
       const scimConfiguration = await this.getScimConfiguration({ id: connection.id });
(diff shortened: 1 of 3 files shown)


src/tools/auth0/handlers/scimHandler.ts
```diff
@@ -125,7 +117,19 @@ export default class ScimHandler {
...
+        log.warn(warningMessage);
+        return { data: null };
+      };
+
+      // Skip the connection if it returns 403. Looks like "scim_config" permissions are not enabled on Management API. 
+			if (errorData?.statusCode === 403) {
+        const warningMessage = `SCIM configuration is not enabled on connection \"${ this.currentConnectionId }\".`;
+        log.warn(warningMessage);
+        return { data: null };
+      }
```
bweisberger, line 132: After this change will we still be able to manage connections via IaC even if we don't use SCIM?

src/tools/auth0/handlers/scimHandler.ts
```diff
@@ -55,19 +56,16 @@ export default class ScimHandler {
...
-				// TODO: However, this logic needs to be re-worked.
-				await sleep(500);
-			} catch (err) {
-				// Skip the connection if it returns 404. This can happen if `SCIM` is not enabled on a `SCIM` connection. 
-				if (err !== 'SCIM_NOT_FOUND') throw err;
-			}
+      
+      // To avoid rate limiter error, we making API requests with a small delay.
+      // TODO: However, this logic needs to be re-worked.
+      await sleep(500);
```
bweisberger, line 62: Why is the SCIM update hitting rate limits when other connection updates are not?

bweisberger (commented):
I hope this fixes [this issue](https://github.com/auth0/auth0-deploy-cli/issues/924). I left a couple questions. Thank you!

src/tools/auth0/handlers/scimHandler.ts
nandan-bhat: Hello @bweisberger,

The main challenge we're facing is that there isn't a straightforward way to get a complete list of SCIM-enabled connections. As a result, we have to manually check each SCIM connection to see if it's enabled or not, which is a limitation of the API.

While going through this process, we encountered rate limits (429 errors) when using the `/v2/connections/get-scim-configuration` endpoint, which led us to introduce `sleep` in the operations.

Another complication is that `auth0-deploy-cli` primarily relies on `node-auth0` for its features. However, the current version of `node-auth0` used in `auth0-deploy-cli` doesn't support SCIM. To work around this, we're directly using `Management API` endpoints to handle SCIM functionalities.

We understand this isn't the most efficient solution, but it's the approach we've had to take for now.

src/tools/auth0/handlers/scimHandler.ts
nandan-bhat: Hello @bweisberger ,

Yes. That's correct !

src/tools/auth0/handlers/scimHandler.ts
bweisberger: Got it, makes sense. I think we're going to stay pinned to the version before 7.23.0 this one for now - we don't use SCIM and it looks like this change would add at least a couple minutes to our deploys. I'll be following along though as you all work through solutions. Thanks!

src/tools/auth0/handlers/scimHandler.ts
bweisberger: Ok, I ask because after looking [here](https://github.com/auth0/auth0-deploy-cli/blob/master/src/tools/auth0/handlers/connections.ts#L102) I got the impression that all connections updates use the scimHandler, which seems to be skipping connections updates [here](https://github.com/auth0/auth0-deploy-cli/pull/925/files#diff-38cb0d98b14b46a15ddcd2a6398c0b97fabd410e28af0e5b10579402e3359c68R66) if SCIM updates fail. I'm sure I'm missing plenty, but just wanted to check

nandan-bhat: If a SCIM operation encounters an error other than `404` or `403`, it could potentially cause the connection to fail. While a `429` error doesn't necessarily mean immediate failure, our goal is to avoid triggering 429 errors altogether if possible. Because, if SCIM causes `429`, it'll fail connection as well.

src/tools/auth0/handlers/scimHandler.ts
nandan-bhat: @bweisberger,
Got it! Yes, SCIM was the only feature introduced in version 7.23.0. If you're not utilizing SCIM functionality, sticking with the previous version makes sense, especially if it helps keep your deploy times shorter by a few minutes.

Thank you !!

gyaneshgouraw-okta (approved):
Looks Ok to me

developerkunal (approved):

PR #952, opened 2024-09-04
Fix preserve keywords in clientGrants audience

kushalshit27:
### 🔧 Changes

When `AUTH0_PRESERVE_KEYWORDS : true`, `export `operation should preserve the mapping for `audience` key.

example:
```
clientGrants:
  - client_id: My Application
    audience: https://##AUTH0_DOMAIN##/api/v2/
```

### 🔬 Testing

Added a new unit test cases to the main keyword preservation test with `clientGrants` resource. 

### 📝 Checklist

### DIFF
--- a/test/context/yaml/clientGrants.test.js
+++ b/test/context/yaml/clientGrants.test.js
@@ -37,6 +37,40 @@ describe('#YAML context client grants', () => {
     expect(context.assets.clientGrants).to.deep.equal(target);
   });
 
+  it('should process client grants with AUTH0_PRESERVE_KEYWORDS: true', async () => {
+    const dir = path.join(testDataDir, 'yaml', 'clientGrants');
+    cleanThenMkdir(dir);
+
+    const yaml = `
+    clientGrants:
+      - client_id: "my_client_id"
+        name: "My M2M"
+        audience: "https://##ENV##.myapp.com/api/v1"
+        scope:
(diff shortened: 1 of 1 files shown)


willvedd (approved):
Code itself looks good to me. Tip: it's easier to update the changelog in this PR.

src/context/yaml/handlers/clientGrants.ts
```diff
@@ -11,13 +12,34 @@ async function parse(context: YAMLContext): Promise<ParsedClientGrants> {
...
+    // here can not use client from context.assets because it does not have the client_id
+    const clients = await context.mgmtClient.clients.getAll({
+      paginate: true,
+      include_totals: true,
+    });
+
+    return {
+      clientGrants: clientGrants.map((grant) => {
+        const dumpGrant = { ...grant };
+        dumpGrant.client_id = convertClientNameToId(dumpGrant.client_id, clients || []);
```
willvedd, line 31: Sorry, I didn't catch this right before 🤦‍♂️ .

We shouldn't be dumping the client ID because it isn't going to be the same on each tenant. Remember that the Deploy CLI is a multi-tenant configuration management tool. The only way to maintain agnostic to tenant is by continuing to use name over ID.

willvedd (changes requested):

Issue #988, opened 2024-11-26
export fails on triggers with error: "cannot list action bindings for an entity-bound trigger"


bhaugeea:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/tag/auth0-deploy-cli) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Given `"triggers"` as an included resource type, Deploy CLI gets all triggers and then for each one gets all bindings. For entity-bound triggers, Auth0 Management API responds with `400 cannot list action bindings for an entity-bound trigger`. Deploy CLI then fails with that error. There is [no logic that I can see](https://github.com/auth0/auth0-deploy-cli/blob/master/src/tools/auth0/handlers/triggers.ts#L65-L67) in Deploy CLI to check whether a trigger is entity-bound before trying to list bindings.

### Expectation

Deploy CLI should gracefully exclude off-limits entities from both export and import. I'm happy to open a PR to check whether the trigger is entity-bound, but I'm not sure if simply excluding that data from the export would create diffing/import problems. Just let me know.

### Reproduction

I am testing in a more or less fresh tenant as a new Auth0 customer, and this is my first time trying to use Deploy CLI. The issue is consistent and can be seen in the code linked in the description and also in the Management API test request described below. I'm not sure what if any customizations could be made in the tenant to avoid the error, and similarly not sure why others aren't encountering/reporting it.

1. Check the response that Auth0 Management API returns for the offending request [here](https://github.com/auth0/auth0-deploy-cli/blob/master/src/tools/auth0/handlers/triggers.ts#L65-L67) with `custom-token-exchange`, which in my case is said by `GET /triggers` to be (the only) entity-bound. It is the error thrown by Deploy CLI.
[...]

PR #990, opened 2024-11-27
Temp fix for entity-bound triggers error

Ryan-Tassat:
### 🔧 Changes

Just a quick fix for the issue introduced with custom-token-exchange being returned when calling the triggers API. 
This will continue even if the API responds with the "cannot list action bindings for an entity-bound trigger" response (400 response code). May be a cleaner way to do this, but it does the job for now!

All it does is adds a new check in the catch section of the try-catch that makes a call to the triggerBindings endpoint.

### 📚 References

- https://github.com/auth0/auth0-deploy-cli/issues/988

### 🔬 Testing

Try running a full export with debug mode enabled. Without this "fix" the export will break and not complete successfully. With this fix, it'll just "warn" about the issue.

### 📝 Checklist

- [ X ] All new/changed/fixed functionality is covered by tests (or N/A)
- [ X ] I have added documentation for all new/changed functionality (or N/A)

### DIFF
--- a/src/tools/auth0/handlers/triggers.ts
+++ b/src/tools/auth0/handlers/triggers.ts
@@ -86,6 +86,11 @@ export default class TriggersHandler extends DefaultHandler {
         return {};
       }
 
+      if (err.message === "cannot list action bindings for an entity-bound trigger") {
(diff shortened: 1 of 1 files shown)


Ryan-Tassat:
Thanks @kushalshit27 - Feel free to bin this PR. The builds didn't pass 😅  - My TS is rusty and I somehow fudged the definition of triggerId

PR #1113, opened 2025-06-16
feat: Support for Multiple Custom Domains

kushalshit27:
### 🔧 Changes

Example:
```diff
customDomains:
  - domain: my_domain.com
    .....................
+    domain_metadata:
+      myKey: value
```
📌 Note:
- `primary` field is deprecated and may be removed in future versions.
- `verification_method` field is deprecated and may be removed in future versions.

### 📚 References

- https://auth0.com/docs/api/management/v2/custom-domains/post-custom-domains
- https://auth0.com/docs/api/management/v2/custom-domains/get-custom-domains

### 🔬 Testing
- Unit Test added

### 📝 Checklist


examples/yaml/tenant.yaml
```diff
@@ -272,3 +272,10 @@ networkACLs:
       scope: 'authentication'
       match:
         geo_country_codes: ['US', 'CA']
+
+customDomains:
+  - domain: my_domain.com
+    type: auth0_managed_certs
+    domain_metadata:
```
ramya18101, line 280: Can we include tls_policy too..?

src/tools/auth0/handlers/customDomains.ts
```diff
@@ -31,10 +43,20 @@ export default class CustomDomainsHadnler extends DefaultAPIHandler {
       ...config,
       type: 'customDomains',
       id: 'custom_domain_id',
-      identifiers: ['domain'],
+      identifiers: ['custom_domain_id', 'domain'],
       stripCreateFields: ['status', 'primary', 'verification'],
+      stripUpdateFields: [
+        'status',
```
ramya18101, line 54: I also see another field name `certificate`(read-only field similar to verification) in tf-provider PR, Any reason on not including in deploy-cli..?

test/tools/auth0/handlers/customDomains.test.ts
```diff
@@ -288,15 +288,266 @@ describe('#customDomains handler', () => {
...
+      status: 'ready',
+    };
+
+    const updatedCustomDomainWithExtraFields = {
+      domain: 'test.example.com', // should be stripped
+      type: 'auth0_managed_certs', // should be stripped
+      status: 'ready', // should be stripped
+      primary: true, // should be stripped
+      verification: { method: 'cname' }, // should be stripped
+      verification_method: 'cname', // should be stripped
```
ramya18101, line 371: Why do we have `verification_method` field separately in domain.. As we already have a field `method` in verifications!

kushalshit27: `verification` and `verification_method` can co-exist in the configuration at the same time. This function tests that they are stripped properly, as per our logic during update.

examples/yaml/tenant.yaml
kushalshit27: Sure

src/tools/auth0/handlers/customDomains.ts
kushalshit27: Sure, this was not added in the previous implementation, but we can add it now.

ramya18101: I saw it being added in the `stripUpdateFields` , but not in the schema

kushalshit27: Yes, because the schema is used for input validation,   `certificate` does not need validation.

examples/yaml/tenant.yaml
kushalshit27: updated

PR #1166, opened 2025-09-25
feat: support UserAttributeProfiles and related properties

ewanharris:
### 🔧 Changes

> [!WARNING]
> This feature is currently behind a feature flag so will not work until the release is finalised

Adds support for the `userAttributeProfiles` resource and the related property on `selfServiceProfiles` to link the two. The `user_attribute_profile_id` also supports setting the `name` as the value which will then be mapped to an idea similar to the `enabled_clients`.

Also includes a fix for the checkpoint pagination helper as not all endpoints will return the total when using this pagination type.

Below is a small example that demonstrates the usage of `userAttributeProfiles` and linking it to a `selfServiceProfile`

```yaml
selfServiceProfiles:
  - name: self-service-profile-2
    allowed_strategies:
      - google-apps
      - okta
    description: test description self-service-profile-2
    user_attribute_profile_id: Test User Attribute
userAttributeProfiles:
  - name: Test User Attribute
    user_attributes:
      email:
        label: Email
        description: User's email address
        oidc_mapping:
          mapping: email
        auth0_mapping: email
        profile_required: true
    user_id:
      oidc_mapping: sub
      saml_mapping:
        - http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier
        - http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn
        - http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name
      scim_mapping: externalId
```

### 📚 References

Requires https://github.com/auth0/node-auth0/pull/1186 to be merged and released first

### 🔬 Testing

Tests added, E2E tests not added yet as can't enable this feature for the tenant currently.

Manual testing performed:
- Export resources
- Import resources
- Create new resources
- Create a new resource where the name is used as `user_attribute_profile_id` to ensure ordering works

### 📝 Checklist


src/context/yaml/handlers/userAttributeProfiles.ts
```diff
@@ -0,0 +1,35 @@
...
+  const { userAttributeProfiles } = context.assets;
+
+  if (!userAttributeProfiles) return { userAttributeProfiles: null };
+
+  return {
+    userAttributeProfiles,
+  };
+}
+
+async function dump(context: YAMLContext): Promise<ParsedUserAttributeProfiles> {
```
ramya18101: Can we have 1 common function named `parseAndDump`

src/tools/auth0/handlers/userAttributeProfiles.ts
```diff
@@ -0,0 +1,244 @@
+import {
+  UserAttributeProfile
```
ramya18101, line 2: Since we set the order for SSP to 60, do we need to ensure that UserAttribute order < 60?”

src/tools/auth0/handlers/selfServiceProfiles.ts
```diff
@@ -81,6 +82,9 @@ export const schema = {
           },
         },
       },
+      user_attribute_profile_id: {
```
ramya18101, line 85: How are we ensuring either `user_attribute_profile_id` or `user_attributes` parses

ramya18101 (changes requested):

src/context/yaml/handlers/userAttributeProfiles.ts
ewanharris: Changed in b4a5ad0caa676cee21f85917a0030fb19711d903

src/tools/auth0/handlers/userAttributeProfiles.ts
ewanharris: Changed in b4a5ad0caa676cee21f85917a0030fb19711d903

src/tools/auth0/handlers/selfServiceProfiles.ts
ewanharris: I think the check here needs to be like `if (user_attribute_profile_id !== "" && user_attributes.length > 0)` as an empty array is allowed for `user_attributes` when `user_attribute_profile_id` is set (the API just doesn't respect this right now.

I don't think that's expressable in jsonschema so I might need to add a codebased check I assume in `processChanges`

ewanharris: Done in c6ead17cea77265af9436e428ab4e94788550c2e

PR #1219, opened 2025-11-29
feat: Implement migration for deprecated client fields in configuration

palashgdev:
### 🔧 Changes

This change specifically addresses backward compatibility for client application settings. We've enhanced the tool to reliably handle the old cross_origin_auth field and smoothly migrate its value to the new cross_origin_authentication field.

### 🔬 Testing

This change is purely a code style/best practice fix within a unit test. It ensures the existing logic for handling deprecated client fields during YAML load remains correct and reliable. No change in core functionality is expected.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/clients.ts
+++ b/src/tools/auth0/handlers/clients.ts
@@ -3,11 +3,13 @@ import {
   ClientExpressConfiguration,
   ClientOrganizationRequireBehaviorEnum,
 } from 'auth0';
+import _ from 'lodash';
 import { Assets, Auth0APIClient } from '../../../types';
 import { paginate } from '../client';
 import DefaultAPIHandler from './default';
 import { getConnectionProfile } from './connectionProfiles';
 import { getUserAttributeProfiles } from './userAttributeProfiles';
+import log from '../../../logger';
 
 const multiResourceRefreshTokenPoliciesSchema = {
   type: ['array', 'null'],
@@ -316,6 +318,11 @@ export default class ClientHandler extends DefaultAPIHandler {
 
     assets.clients = await this.sanitizeMapExpressConfiguration(this.client, clients);
 
+    assets.clients = this.normalizeClientFields({
+      clients,
+      fields: [{ newField: 'cross_origin_authentication', deprecatedField: 'cross_origin_auth' }],
+    });
+
     const excludedClients = (assets.exclude && assets.exclude.clients) || [];
 
     const { del, update, create, conflicts } = await this.calcChanges(assets);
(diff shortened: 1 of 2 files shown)


src/tools/auth0/handlers/clients.ts
```diff
@@ -3,11 +3,13 @@ import {
   ClientExpressConfiguration,
   ClientOrganizationRequireBehaviorEnum,
 } from 'auth0';
+import _ from 'lodash';
```
kushalshit27, line 6: Please import functions instead of `_ ` for lodash.

src/tools/auth0/handlers/clients.ts
```diff
@@ -373,10 +380,56 @@ export default class ClientHandler extends DefaultAPIHandler {
       is_global: false,
     });
 
-    this.existing = clients;
+    const sanitizedClients = this.normalizeClientFields({
+      clients,
+      fields: [{ newField: 'cross_origin_authentication', deprecatedField: 'cross_origin_auth' }],
+    });
```
kushalshit27, line 386: move this logic under existing  to

```
// Sanitize client fields
    const sanitizeClientFields = (list: Client[]): Client[] => ....
```

src/tools/auth0/handlers/clients.ts
```diff
@@ -373,10 +380,56 @@ export default class ClientHandler extends DefaultAPIHandler {
...
     return this.existing;
   }
 
+  /**
+   * @description Maps deprecated client fields to their new counterparts and removes the deprecated field.
+   * If a deprecated field exists, its value is always used for the new field, ensuring data migration
+   * and preventing loss of configuration data during schema transitions.
+   * @returns Client[]
+   */
+  normalizeClientFields = ({
```
kushalshit27, line 398: There is already a function named `sanitizeClientFields`  . please be specific with the function name

src/tools/auth0/handlers/clients.ts
```diff
@@ -373,10 +380,56 @@ export default class ClientHandler extends DefaultAPIHandler {
...
+    });
+
+    this.existing = sanitizedClients;
     return this.existing;
   }
 
+  /**
+   * @description Maps deprecated client fields to their new counterparts and removes the deprecated field.
+   * If a deprecated field exists, its value is always used for the new field, ensuring data migration
+   * and preventing loss of configuration data during schema transitions.
```
kushalshit27, line 395: Logic is not correctly handled. 

- If only `cross_origin_auth` exists, use that value. (with a warn log)

- if both exists
```
	cross_origin_auth: false
	cross_origin_authentication: false  
```
Use the value of `cross_origin_authentication` (with a warn log for cross_origin_auth)


- if both exists
```
	cross_origin_auth: true
	cross_origin_authentication: false  
```
Use the value of `cross_origin_authentication` (with a warn log for cross_origin_auth)

Always give preference to the new key `cross_origin_authentication`.

src/tools/auth0/handlers/clients.ts
```diff
@@ -373,10 +380,56 @@ export default class ClientHandler extends DefaultAPIHandler {
...
+              }
+              // Remove the deprecated field
+              return _.omit(acc, deprecatedField);
+            }
+
+            return acc;
+          },
+          { ...client } as Record<string, unknown>
+        ) as Client
+    );
```
kushalshit27, line 431: reduce complexity, read more on [link](https://en.wikipedia.org/wiki/KISS_principle)

palashgdev:
closing in favor of https://github.com/auth0/auth0-deploy-cli/pull/1223

PR #1207, opened 2025-11-18
Upgrading node-auth0 from v4 to v5 [beta]

kushalshit27:
### 🔧 Changes

Upgrading node-auth0 from v4 to v5. This is a necessary upgrade because all new Auth0 features and dependency patches will be released on v5 in future.

```
Requires Node.js ^20.19.0, ^22.12.0, or ^24.0.0.
```
#### :bookmark: Note:
This is an internal package update that should not impact any existing operations. 

📚 References
https://github.com/auth0/node-auth0/releases/tag/v5.0.0

### 🔬 Testing
Deploy CLI commands  like import, export 

### 📝 Checklist

- [x] All new/changed/fixed functionality is covered by tests (or N/A)
- [ ] I have added documentation for all new/changed functionality (or N/A)

PR #1222, opened 2025-12-02
8.20.4-beta.0

kushalshit27:
### Changed

- Upgrading node-auth0 from v4 to v5 [beta]. [#1207]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [8.20.4-beta.0] - 2025-12-02
+
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [8.20.4-beta.0] - 2025-12-02
```
ramya18101, line 10: Since the latest release we had `8.21.0` ..
and why patch release, Can we have our release as `[8.22.0-beta.0]`

kushalshit27: Connection Profiles and Express Configuration on Clients are not available in the current v5, that's why `8.21.1-beta.0` is not possible.

PR #1330, opened 2026-03-23
feat: Add support for default custom domain and deprecation warning for enable_custom_domain_in_emails 

ankita10119:
### Context                                                                                                                                                                     
                                                                                                                                                                              
The Auth0 Management API is introducing a new default custom domain concept as part of the Multiple Custom Domains (MCD) GA release. Tenants can now designate one of their verified custom domains as the default, this domain is used in email/phone notification links (e.g. password reset, email verification) when no explicit `auth0-custom-domain `header is sent by the API client.                                                                                                                       
                                                                                                                                                                              
  Two new Management API endpoints are being introduced:
  - `GET /api/v2/custom-domains/default` , returns the current default domain
  - `PATCH /api/v2/custom-domains/default` , sets the default domain

  This PR implements Deploy CLI support for both changes ahead of the API going live.

### 🔧 Changes

`src/tools/auth0/handlers/customDomains.ts`

  - Added `is_default` as an optional boolean field to the custom domain JSON schema so it can be declared in YAML configs
  - `is_default` remains in `stripCreateFields` and `stripUpdateFields`, the regular `PATCH /api/v2/custom-domains/{id} `endpoint rejects it as read-only per the RFC, it can only be set via the dedicated default endpoint
  - Added a `setDefaultDomain()` private method that calls `PATCH /api/v2/custom-domains/default` directly since this endpoint is not yet available in the auth0 Node.js SDK. The method reuses the SDK client's internal auth provider (`_options.authProvider`) via type assertion to avoid duplicating token management logic
  - `processChanges` now calls `setDefaultDomain()` after the normal **create/update/delete** cycle completes, if any domain in the config has `is_default: true`

  `src/tools/auth0/handlers/tenant.ts`

  - Added a deprecation warning when `enable_custom_domain_in_emails` is found in tenant flags, guiding users to migrate to the `is_default` field on customDomains
  - The flag remains in `allowedTenantFlags` since it hasn't been removed from the API yet

  `examples/yaml/tenant.yaml`

  - Added `is_default: true` to the custom domain example to document the new field

 ` test/tools/auth0/handlers/customDomains.test.ts`

  - Updated the existing strip-fields test to mock _options on the custom domains client and stub `globalThis.fetch`, since `is_default: true` in the test data now triggers the `setDefaultDomain` call

### 📚 References



### 🔬 Testing

#### Unit tests
  - All 15 existing custom domain handler tests pass
  - Verified `is_default` is correctly stripped from both create and update API payloads
  - Verified `setDefaultDomain` is triggered when `is_default: true `is present in the config

####  Manual testing against a real tenant                                                                                                                                        
  - Confirmed the Auth0 Management API already returns `is_default` on `GET /api/v2/custom-domains` responses - the field is live                                                                                                                
  - Ran the local build with a YAML declaring a custom domain as `is_default: true`                                                                                           
  - All custom domains were updated successfully via the regular create/update path                                                                                           
  - The `setDefault` SDK call successfully reached `PATCH /api/v2/custom-domains/default` with the correct payload
  - Confirmed via `GET /api/v2/custom-domains` after the import that the target domain
    was set as the default - full end-to-end flow verified  

### 📝 Checklist

### DIFF
--- a/test/tools/auth0/handlers/connections.tests.js
+++ b/test/tools/auth0/handlers/connections.tests.js
@@ -4,7 +4,6 @@ import pageClient from '../../../../src/tools/auth0/client';
 const Ajv = require('ajv');
 const { expect } = require('chai');
 const sinon = require('sinon');
-const { Management } = require('auth0');
 const connections = require('../../../../src/tools/auth0/handlers/connections');
 const utils = require('../../../../src/tools/utils');
 const { mockPagedData } = require('../../../utils');


src/tools/auth0/handlers/customDomains.ts
```diff
@@ -40,6 +40,11 @@ export const schema = {
         description:
           'Relying Party ID (rpId) to be used for Passkeys on this custom domain. If not provided or set to null, the full domain will be used.',
       },
+      is_default: {
+        type: 'boolean',
+        description:
+          'Whether this custom domain is the default domain used for email notifications. Set via PATCH /api/v2/custom-domains/default.',
```
kushalshit27:
```suggestion
          'Whether this custom domain is the default domain used for email notifications.',
```

src/tools/auth0/handlers/customDomains.ts
```diff
@@ -128,5 +133,37 @@ export default class CustomDomainsHadnler extends DefaultAPIHandler {
...
+
+    const response = await fetch(`${baseUrl}/api/v2/custom-domains/default`, {
+      method: 'PATCH',
+      headers: {
+        ...authRequest.headers,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ domain }),
+    });
+
```
kushalshit27: always use auth0 sdk

src/tools/auth0/handlers/customDomains.ts
```diff
@@ -128,5 +133,37 @@ export default class CustomDomainsHadnler extends DefaultAPIHandler {
...
+    // Set default domain via PATCH /api/v2/custom-domains/default if declared in assets.
+    // This endpoint is not yet in the auth0 SDK, so we call it directly.
+    const defaultDomain = customDomains.find((d) => d.is_default === true);
+    if (defaultDomain) {
+      await this.setDefaultDomain(defaultDomain.domain);
+      log.info(`Set default custom domain: ${defaultDomain.domain}`);
+    }
+  }
+
+  private async setDefaultDomain(domain: string): Promise<void> {
```
kushalshit27: this function can be replaced with node SDK function

src/tools/auth0/handlers/tenant.ts
```diff
@@ -195,6 +195,13 @@ export default class TenantHandler extends DefaultHandler {
       }
     }
 
+    if ((tenant.flags as Record<string, unknown>)?.enable_custom_domain_in_emails !== undefined) {
+      log.warn(
+        'The "enable_custom_domain_in_emails" tenant flag is deprecated. ' +
+          'Use the "is_default" field on customDomains to configure the default domain instead.'
+      );
+    }
```
kushalshit27, line 204: is it needed?

test/tools/auth0/handlers/customDomains.test.ts
```diff
@@ -391,13 +397,22 @@ describe('#customDomains handler', () => {
       }),
     };
 
+    // Mock global fetch for the setDefaultDomain call
+    const originalFetch = globalThis.fetch;
```
kushalshit27: No : const originalFetch = globalThis.fetch;

 always use SDK

src/tools/auth0/handlers/customDomains.ts
ankita10119: Addressed and updated

src/tools/auth0/handlers/customDomains.ts
ankita10119: Addressed and updated

src/tools/auth0/handlers/tenant.ts
ankita10119: Yes, the RFC explicitly deprecates `enable_custom_domain_in_emails` in favor of `is_default`, and users need migration guidance

test/tools/auth0/handlers/customDomains.test.ts
ankita10119: Addressed and updated

src/tools/auth0/handlers/customDomains.ts
ankita10119: Addressed and updated

src/tools/auth0/handlers/customDomains.ts
```diff
@@ -128,5 +133,12 @@ export default class CustomDomainsHadnler extends DefaultAPIHandler {
 
     const changes = await this.calcChanges(assets);
     await super.processChanges(assets, changes);
+
+    // If a domain is marked as is_default, set it as the tenant's default custom domain.
+    const defaultDomain = customDomains.find((d) => d.is_default === true);
+    if (defaultDomain) {
+      await this.client.customDomains.setDefault({ domain: defaultDomain.domain });
```
kushalshit27: What if this setDefault returns error !

src/tools/auth0/handlers/tenant.ts
kushalshit27: Will this flag be removed by `removeUnallowedTenantFlags`?

ankita10119: Addressed and updated

Issue #1186, opened 2025-10-24
Sorted object dumps


RenoMeyer:
### Checklist

- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

Hey guys, I noticed that some of the outputs from your export command are sorted in various manners.

For example the `custom-text.json` is sorted based on the `enabled_locales` in the `tenant.json`. As I have a multi-tenant project with different default locales, the exports differ depending on which tenant I use (while the translations should be the same except some keyword replacements).

Another example is that action secrets which are listed in the order they've been added/edited. Since this is a manual process (as far as I understand), the order can easily differ from tenant to tenant.

### Describe the ideal solution

I think both these examples could be easily solved by running an alphabetical sort over the json files before outputting from the export. The `partial.json` already seem to follow such a pattern if I'm not mistaken.

Thanks for your consideration and your awesome product!

kushalshit27:
Thanks for the feature request!
Great suggestion! We will evaluate how to incorporate it. 🤝

PR #1335, opened 2026-03-25
feat: add --export_ordered flag to produce stable JSON exports

zariok:
Auth0's API returns resource attributes in inconsistent order, causing noisy diffs in version control even when nothing has changed.

### 🔧 Changes

Adds `--export-ordered` / `-s` CLI flag and `AUTH0_EXPORT_ORDERED` config key to sort all exported JSON and YAML keys alphabetically at write time, producing output equivalent to `jq --sort-keys` — no external tooling required.

Implemented in `dumpJSON()` — the shared write utility used by all directory-format handlers — so every resource type gets the behavior automatically with no call-site changes. Also fixes two handlers (`actions.ts`, `triggers.ts`) that were bypassing `dumpJSON` entirely with raw `fs.writeFileSync` calls.

### 📚 References

Resolves: #1186

### 🔬 Testing

- [ ] `npm run build` passes with no TypeScript errors
- [ ] `npm test` passes
- [ ] Export with `--export-ordered` produces files where `diff <(jq --sort-keys . file.json) file.json` is empty
- [ ] Export without `--export-ordered` is unchanged from current behavior
- [ ] `AUTH0_EXPORT_ORDERED=true` in config file produces the same result as the CLI flag

### 📝 Checklist


src/commands/export.ts
```diff
@@ -45,6 +46,11 @@ export default async function exportCMD(params: ExportParams) {
     overrides.AUTH0_EXPORT_IDENTIFIERS = exportIds;
   }
 
+  // Allow passed in sort_keys to override the configured one
+  if (sortKeys) {
+    overrides.AUTH0_SORT_JSON_KEYS = sortKeys;
```
kushalshit27: `AUTH0_EXPORT_ORDERED` would be a better name.

src/args.ts
```diff
@@ -110,6 +111,12 @@ function getParams(): CliParams {
         type: 'boolean',
         default: false,
       },
+      sort_keys: {
+        alias: 's',
+        describe: 'Sort keys in exported JSON files for consistent diffs.',
+        type: 'boolean',
+        default: false,
+      },
```
kushalshit27:
```js
  export_ordered: {                                                                                                                             
    alias: 's',                                                                                                                                 
    describe: 'Order keys alphabetically in exported files for consistent diffs.',                                                              
    type: 'boolean',                                                                                                                            
    default: false,                                                                                                                             
  }
```

src/types.ts
```diff
@@ -77,6 +77,7 @@ export type Config = {
   AUTH0_RETRY_MAX_DELAY_MS?: number;
   AUTH0_KEYWORD_REPLACE_MAPPINGS?: KeywordMappings;
   AUTH0_EXPORT_IDENTIFIERS?: boolean;
+  AUTH0_SORT_JSON_KEYS?: boolean;
```
kushalshit27: AUTH0_EXPORT_ORDERED would be a better name.

src/utils.ts
```diff
@@ -51,10 +52,27 @@ export function loadJSON(
   }
 }
 
+function sortedKeysReplacer(_key: string, value: unknown): unknown {
```
kushalshit27: Try to reduce the function complexity 
- Redundant cast inside reduce
-  Double-casting value
- Can be simplify with Object.fromEntries

docs/using-as-cli.md
```diff
@@ -37,6 +37,10 @@ Boolean. Enables more verbose error logging; useful during troubleshooting. Defa
 
 A url for proxying requests. Only set this if you are behind a proxy.
 
+### `--sort-keys`
```
kushalshit27: This should be `export_ordered`

zariok: Combed through and renamed other references to sortKeys as well

src/context/yaml/index.ts
```diff
@@ -215,7 +215,8 @@ export default class YAMLContext {
     }
 
     // Write YAML File
-    const raw = yaml.dump(cleaned);
+    const exportOrdered = Boolean(this.config.AUTH0_EXPORT_ORDERED);
+    const raw = yaml.dump(cleaned, { exportOrdered });
```
kushalshit27:
```suggestion
    const raw = yaml.dump(cleaned, { sortKeys: exportOrdered });
```

src/args.ts
```diff
@@ -110,6 +111,12 @@ function getParams(): CliParams {
         type: 'boolean',
         default: false,
       },
+      export_ordered: {
+        alias: 's',
+        describe: 'Order keys in exported JSON files for consistent diffs.',
+        type: 'boolean',
+        default: false,
```
kushalshit27: Please remove the default value here. We need to distinguish between “flag not provided” and an explicit false; otherwise CLI override behavior for `export_ordered` can be incorrect.

PR #1354, opened 2026-04-07
fix: resolve silent process exit when deploying 3+ organizations with discovery domains

ankita10119:
When a tenant configuration includes 3 or more organizations that each have `discovery_domains`, running `a0deploy import` silently exits mid-run without error. No discovery domains are created or updated, and the "Import Successful" message never appears.

This affects both first-time deploys (create path) and subsequent deploys (update path). The process exits after the delete-warning phase, before any organization changes are applied.

### 🔧 Changes

 - organizations.ts: Discovery domain create/update/delete operations no longer route through the shared request pool. They now use Promise.all and direct API calls, consistent with how connections and `client_grants` sub-operations are handled in the same handler.
  - connections.ts: Fixed a crash at module load time when ConnectionDpopSigningAlgEnum is not present in the installed auth0 SDK version. The schema validation for `dpop_signing_alg` now gracefully omits the enum constraint when the SDK does not provide it.

  **Backward Compatibility**

  No breaking changes. Behavior is identical for:

  - Tenants with fewer than 3 organizations with discovery domains (the deadlock condition was never triggered).
  - Tenants without the Organization Discovery Domains feature enabled.
  - All other resource types — only the discovery domain sub-operations inside OrganizationsHandler are affected.

### 🔬 Testing

-  Added a regression test that reproduces the deadlock condition directly
- Tested with real tenant data as well

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/organizations.ts
+++ b/src/tools/auth0/handlers/organizations.ts
@@ -167,12 +167,9 @@ export default class OrganizationsHandler extends DefaultHandler {
     }
 
     if (typeof org.discovery_domains !== 'undefined' && org.discovery_domains.length > 0) {
-      await this.client.pool
-        .addEachTask({
-          data: org.discovery_domains,
-          generator: (
-            discoveryDomain: Management.CreateOrganizationDiscoveryDomainRequestContent
-          ) =>
+      await Promise.all(
+        org.discovery_domains.map(
+          (discoveryDomain: Management.CreateOrganizationDiscoveryDomainRequestContent) =>
             this.createOrganizationDiscoveryDomain(created.id, {
               domain: discoveryDomain?.domain,
               status: discoveryDomain?.status,
@@ -181,9 +178,9 @@ export default class OrganizationsHandler extends DefaultHandler {
               throw new Error(
                 `Problem creating discovery domain ${discoveryDomain?.domain} for organization ${created.id}\n${err}`
               );
-            }),
-        })
-        .promise();
+            })
+        )
+      );
     }
     return created;
   }
@@ -664,15 +661,10 @@ export default class OrganizationsHandler extends DefaultHandler {
     log.debug(
       `Creating discovery domain ${discoveryDomain.domain} for organization ${organizationId}`
     );
-    const orgDiscoveryDomain = await this.client.pool
-      .addSingleTask({
-        data: {
-          id: organizationId,
-        },
-        generator: (args) =>
-          this.client.organizations.discoveryDomains.create(args.id, discoveryDomain),
-      })
-      .promise();
+    const orgDiscoveryDomain = await this.client.organizations.discoveryDomains.create(
+      organizationId,
+      discoveryDomain
+    );
     return orgDiscoveryDomain;
   }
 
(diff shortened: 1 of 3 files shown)


src/tools/auth0/handlers/organizations.ts
```diff
@@ -167,12 +167,9 @@ export default class OrganizationsHandler extends DefaultHandler {
     }
 
     if (typeof org.discovery_domains !== 'undefined' && org.discovery_domains.length > 0) {
-      await this.client.pool
-        .addEachTask({
-          data: org.discovery_domains,
-          generator: (
-            discoveryDomain: Management.CreateOrganizationDiscoveryDomainRequestContent
-          ) =>
+      await Promise.all(
```
kushalshit27, line 170: always prefer to use `this.client.pool`

src/tools/auth0/handlers/organizations.ts
```diff
@@ -714,15 +701,6 @@ export default class OrganizationsHandler extends DefaultHandler {
     discoveryDomainId: string
   ): Promise<void> {
     log.debug(`Deleting discovery domain ${discoveryDomain} for organization ${organizationId}`);
-    await this.client.pool
```
kushalshit27, line 717: this.client.pool.addSingleTask should not create issue

src/tools/auth0/handlers/organizations.ts
```diff
@@ -692,19 +684,14 @@ export default class OrganizationsHandler extends DefaultHandler {
       )}`
     );
 
-    const discoveryDomainUpdated = await this.client.pool
```
kushalshit27, line 695: this.client.pool.addSingleTask should not create issue

kushalshit27 (commented):
Please verify,   ``this.client.pool `` and `` this.client.pool.addSingleTask`` should not create an issue; issue should be somewhere else.

src/tools/auth0/handlers/organizations.ts
ankita10119: The root cause of the deadlock is specifically nested pool usage. `createOrganizations()` runs all org tasks through      `this.client.pool.addEachTask()` with the pool configured at `concurrencyLimit: 3`. When 3 orgs are processed concurrently, all 3 slots are occupied. 
  If `createOrganization()` then calls addEachTask on the same pool for discovery domains, those inner tasks queue up but can never start, freeSlots = 0, while the 
outer tasks are blocked waiting for them. Classic deadlock.                                                                                                     
                                                                                                                                                                  Going back to `this.client.pool `here would reintroduce the bug. Also, connections  and client_grants  sub-operations in the exact same `createOrganization()` function already use Promise.all + direct calls, this fix makes discovery_domains consistent with that established pattern

src/tools/auth0/handlers/organizations.ts
ankita10119: `addSingleTask` and `addEachTask` go through the exact same addGenericTask path in the pool and are both subject to the same pool level `concurrencyLimit` .freeSlots is computed as concurrencyLimit - activePromiseCount , and `_triggerNow()` only starts a task when freeSlots > 0.

`updateOrganizations()` also uses `this.client.pool.addEachTask()` as the outer loop. With 3+ orgs being updated concurrently, activePromiseCount = 3 = concurrencyLimit, leaving freeSlots = 0. An inner addSingleTask would queue up but `_triggerNow()` would never fire because the outer tasks never complete, they are waiting for the inner tasks. Same deadlock, different method.

The direct API call pattern is also consistent with how the other sub-operations (connections, client_grants) are handled throughout this handler

kushalshit27 (dismissed):

kushalshit27 (approved):

PR #1360, opened 2026-04-13
fix: strip unresolved keyword placeholders from API payloads before deploy                                                                                      

ankita10119:
### 🔧 Changes

When a keyword placeholder (e.g. `##MY_SECRET##`) is not resolved, because the corresponding environment variable is absent or the keyword file is missing, the raw `##...##` string was previously forwarded to the Auth0 Management API, silently overwriting real tenant values with literal placeholder strings.

  **New utility: `stripUnresolvedPlaceholders` (`src/tools/utils.ts`)**
  - Recursively scans an asset object and removes any string field whose value matches the `##...##` pattern.
  - Emits a `log.warn` for each stripped field, including the full dotted field path, resource type, resource name, and the offending placeholder value, so operators have a clear audit trail.

  **`src/tools/auth0/handlers/default.ts`**
  - `stripUnresolvedPlaceholders` is now applied in all three create/update payload construction paths, chained after `stripObfuscatedFieldsFromPayload`.
  - Fields with unresolved placeholders are dropped from the payload; the existing tenant value is preserved rather than overwritten.

### 📚 References

closes #1344 

### 🔬 Testing

Unit tests covering:
  - A field containing `##PLACEHOLDER##` is stripped and a warning is logged.
  - A nested field (e.g. `options.secret`) containing an unresolved placeholder is stripped with the correct dotted path in the warning.
  - Fields with resolved values pass through unchanged.
  - `null` inputs and array values are returned as-is without modification.
  - Both create and update code paths strip unresolved placeholders before the API call is made.

### 📝 Checklist


src/tools/auth0/handlers/default.ts
```diff
@@ -348,7 +349,15 @@ export default class APIHandler {
...
-              return stripObfuscatedFieldsFromPayload(data, this.sensitiveFieldsToObfuscate);
+              const stripped = stripObfuscatedFieldsFromPayload(
+                data,
+                this.sensitiveFieldsToObfuscate
+              );
+              return stripUnresolvedPlaceholders(
+                stripped as Asset,
+                this.type,
+                this.objString(updateItem)
+              );
```
kushalshit27, line 360: Stripping Unresolved Placeholders is unsafe and goes unnoticed. 

If there are unresolved Placeholders, show an error and exit the flow.

kushalshit27 (changes requested):

src/tools/auth0/handlers/default.ts
ankita10119: Thanks for pointing it out @kushalshit27 
We did consider the fail-fast approach, throwing a validation error when any unresolved placeholder is found. However, the primary use case driving this fix is the export → re-import round trip, where a user exports their tenant (secrets get obfuscated), stores the config in a repo, and re-imports in an environment where not all env vars are available. In that workflow, hard-failing on an unresolved placeholder would block the entire deploy even for resources completely unrelated to the missing secret, which is more disruptive than the problem being solved.                                                                                  
   
The chosen approach follows the principle: "if you don't know the real value, leave the existing tenant value untouched", which is what users naturally expect from a round-trip.                                                                                                                                              
                                         
That said, the "goes unnoticed" concern is fair. The current implementation logs a warn per field with the full path, resource type, resource name, and placeholder value, but we can make this harder to miss by surfacing a summary at the end of the deploy listing all skipped fields. Happy to add that if it addresses the visibility concern.

kushalshit27: This is a good, but fail-fast, approach that would be better in this case, as it makes the user aware of the key used. 

In a multi-tenant workflow, it "goes unnoticed" can be unsafe. I see value in a fail-fast approach.

ankita10119: You're right that fail fast is safer in a misconfiguration scenario, but I want to make sure the fix actually addresses the reported issue, where users export their tenant (secrets become ##...## placeholders) and re import in an environment where not all env vars are set. In that workflow, they're not trying to change those secrets, they just want other config deployed. Fail fast would block the entire deploy every time, which could be more disruptive than the original bug                                                                                                                               
                                                                                                                                                                       
That said, your concern about "goes unnoticed" is valid. Rather than a per field **log.warn** buried in output, I can surface a post-deploy summary using **log.error** listing every skipped field, making it impossible to miss without blocking the deploy. Would that address the visibility concern?

src/tools/utils.ts
```diff
@@ -249,6 +249,36 @@ export const obfuscateSensitiveValues = (
   return newAsset;
 };
 
+const UNRESOLVED_PLACEHOLDER_REGEX = /^##.+##$/;
```
kushalshit27: there is another type ``@@ ...@@``` , please check https://github.com/auth0/auth0-deploy-cli/blob/master/docs/keyword-replacement.md

ankita10119: Addressed and updated

src/tools/auth0/handlers/default.ts
kushalshit27: Removing keys will not help much, as it may trigger a validation error from the API (in most cases), and the deploy will fail halfway.  

That way, failing fast is safer compared to failing later.

kushalshit27 (approved):

src/tools/auth0/handlers/default.ts
rnicholus: this is a mess. the pr description now lists a method that doesn't exist. and deploys fail due to placeholders added by auth0 itself!

Issue #1353, opened 2026-04-06
Organization import with connections crashes with TypeError after v8.23.0   (node-auth0 v4→v5 upgrade)


ZameerIAM:
### Checklist

- [x] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [x] I have searched the [Auth0 Community](https://community.auth0.com/tag/auth0-deploy-cli) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Organization import with enabled connections crashes with TypeError: Cannot read properties of undefined (reading 'id') on versions 8.23.0 through 8.30.0. Works correctly on v8.20.0.                                                                   
                                                            
  **Error**

  Problem running command import during stage processChanges when processing type organizations                                                                           
  Problem creating organizations {"name":"example-org","display_name":"Example Org","metadata":{"org_type":"Rep"},"connections":[{"is_signup_enabled":true,"assign_membership_on_login":true,"show_as_button":true,"connection_id":"con_xxxxx"}],"client_grants":[]}
TypeError: Cannot read properties of undefined (reading 'id')                           
                                                                                          
  **Environment**
                                                                                          
  - Node.js: v23.6.1                                                                      
  - OS: macOS Darwin 25.4.0
  - Export format: directory                                                              
  - M2M app: 251 scopes granted (all org + connection scopes confirmed — not a permissions issue)                                                                                 
   
  **Org file structure (minimal reproduction)**                                               
                                                            
  {                                                                                       
    "name": "example-org",                                  
    "display_name": "Example Org",
    "metadata": {"org_type": "Rep"},
    "connections": [{
      "name": "My-Database-Connection",
      "is_signup_enabled": true,                                                          
      "assign_membership_on_login": true,
      "show_as_button": true                                                              
    }],                                                                                   
    "client_grants": []
  }                                                                                       
                                                            
[...]

PR #1365, opened 2026-04-17
fix: organization creation response handling and add ID validation

kushalshit27:
### 🔧 Changes

 *Root cause*: 
 After the node-auth0 v4→v5 upgrade, `client.organizations.create()` changed from returning { data: org } to returning org directly. The organizations handler was the only one still destructuring { data: created }, causing created to be undefined — hence TypeError: Cannot read properties of undefined (reading 'id') when it tried to use created.id to add connections.
 
 *Fixed*:
- `OrganizationsHandler.createOrganization()` to correctly extract the created organization from the API response. The client method now returns the organization object directly instead of a nested `{ data }` structure.
- Added error handling to validate that the organization ID is present in the response before proceeding with connection/grant associations. If the ID is missing, an error is logged and the response is returned early to prevent downstream failures.

### 📚 References

-  #1353

### 🔬 Testing

- Existing unit tests in `test/tools/auth0/handlers/organizations.tests.js` have been updated to verify the corrected response structure from the organizations client.
- Three test cases now mock the organizations.create() response as the unwrapped data object, consistent with the implementation change.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/organizations.ts
+++ b/src/tools/auth0/handlers/organizations.ts
@@ -145,12 +145,21 @@ export default class OrganizationsHandler extends DefaultHandler {
       delete organization.discovery_domains;
     }
 
-    const { data: created } = await this.client.organizations.create(organization);
+    const created = await this.client.organizations.create(organization);
+
+    if (!created.id) {
+      log.error(
+        `Organization "${organization.name}" was created but the response did not include an ID. Skipping connection/grant association.`
+      );
+      return created;
+    }
+
+    const createdId = created.id;
 
     if (typeof org.connections !== 'undefined' && org.connections.length > 0) {
       await Promise.all(
         org.connections.map((conn) =>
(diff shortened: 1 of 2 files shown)


src/tools/auth0/handlers/organizations.ts
```diff
@@ -145,12 +145,21 @@ export default class OrganizationsHandler extends DefaultHandler {
       delete organization.discovery_domains;
     }
 
-    const { data: created } = await this.client.organizations.create(organization);
+    const created = await this.client.organizations.create(organization);
+
+    if (!created.id) {
+      log.error(
+        `Organization "${organization.name}" was created but the response did not include an ID. Skipping connection/grant association.`
```
ankita10119, line 152: Instead of silently returning here, should we not throw the error?
As the current behaviour creates partial state(org exists, connection missing) with no error propagation

ankita10119 (changes requested):

ankita10119 (approved):

PR #1372, opened 2026-04-28
feat: add support for Third Party Apps CORE EA Release 1 properties

ankita10119:
### 🔧 Changes

Adds deploy CLI support for the new Third Party Apps - CORE EA Release 1 Management API properties. Without these changes, any tenant with third-party client configuration would fail schema validation on import, and users would be unable to manage these settings through the CLI. 

### 📝 Checklist


src/tools/auth0/handlers/clientGrants.ts
```diff
@@ -51,11 +57,17 @@ export default class ClientGrantsHandler extends DefaultHandler {
...
-      stripUpdateFields: ['audience', 'client_id', 'subject_type', 'is_system'],
+      // Nested arrays are not reflected in the type but are supported at runtime.
+      // Try ['client_id', 'audience', 'subject_type'] first, then ['client_id', 'audience']
+      // for regular grants, then ['default_for', 'audience'] for default third-party grants
+      // which have no client_id.
+      identifiers: [
+        'id',
+        ['client_id', 'audience', 'subject_type'],
+        ['client_id', 'audience'],
+        ['default_for', 'audience'],
```
kushalshit27, line 68: Will there be any case where client_id is not present, so we need ['default_for', 'audience']?

ankita10119: Yes, this case exists. `default_for` grants are a new grant type from the **Third Party Apps** feature where `client_id` and `default_for` are mutually exclusive at the API level, a `default_for` grant never has a `client_id`. Since `calculateChanges` requires all fields in a composite identifier to be truthy, `['client_id', 'audience']` skips these grants entirely, so `['default_for', 'audience']` is needed to match them correctly. Added a test to cover this.

kushalshit27 (approved):

PR #1375, opened 2026-05-04
feat: Add support for additional signing algs in OIDC and Okta connections

ramya18101:
### 🔧 Changes

Added support for the following fields under `connections[].options` for strategies `oidc` and `okta`: `token_endpoint_auth_signing_alg`, `id_token_signed_response_algs`, and `token_endpoint_jwtca_aud_format` with appropriate allowed values

### 📚 References

https://auth0.com/docs/api/management/v2/connections/patch-connections-by-id
https://auth0.com/docs/api/management/v2/connections/post-connections

### 🔬 Testing

Added tests to verify that the new options are accepted for `oidc` and `okta` strategies

### 📝 Checklist


src/tools/auth0/handlers/connections.ts
```diff
@@ -14,17 +14,38 @@ import { ConfigFunction } from '../../../configFactory';
...
     dpop_signing_alg: {
       type: 'string',
     },
+    token_endpoint_auth_signing_alg: {
+      type: 'string',
+      enum: ['RS256', 'RS384', 'RS512', 'PS256', 'PS384', 'ES256', 'ES384'],
+    },
+    id_token_signed_response_algs: {
+      type: 'array',
+      items: { type: 'string' },
```
ankita10119: I see at line 29 - `token_endpoint_auth_signing_alg` strictly allow 7 alg, but `id_token_signed_response_algs` accepts any string  
Does Auth0's API enforce a specific set of valid algorithm values here? If so, add an enum to items so invalid values are caught locally instead of returning a cryptic API error

src/tools/auth0/handlers/connections.ts
```diff
@@ -14,17 +14,38 @@ import { ConfigFunction } from '../../../configFactory';
...
 const connectionOptionsSchema = {
   type: 'object',
+  additionalProperties: true,
   properties: {
     dpop_signing_alg: {
       type: 'string',
     },
+    token_endpoint_auth_signing_alg: {
+      type: 'string',
+      enum: ['RS256', 'RS384', 'RS512', 'PS256', 'PS384', 'ES256', 'ES384'],
```
ankita10119: Should we not include PS512 and ES512 here?

src/tools/auth0/handlers/connections.ts
```diff
@@ -14,17 +14,38 @@ import { ConfigFunction } from '../../../configFactory';
 import { paginate } from '../client';
 import ScimHandler from './scimHandler';
 import log from '../../../logger';
+import ValidationError from '../../validationError';
 import { Client } from './clients';
 
 const connectionOptionsSchema = {
   type: 'object',
+  additionalProperties: true,
```
ankita10119, line 22: Is it required here?

src/tools/auth0/handlers/connections.ts
```diff
@@ -398,6 +419,38 @@ export default class ConnectionsHandler extends DefaultAPIHandler {
...
+      const unsupportedFields = oidcOktaOnlyConnectionOptionFields.filter(
+        (field) => field in connection.options
+      );
+
+      if (unsupportedFields.length > 0) {
+        throw new ValidationError(
+          `Connection "${connection.name}": option(s) ${unsupportedFields
+            .map((field) => `"${field}"`)
+            .join(', ')} are only supported for strategies "oidc" and "okta". Found strategy "${
+            strategy || 'unknown'
```
ankita10119: `strategy` is already normalized to '' on line 433 (`connection?.strategy || ''`), but then line 445 does `strategy || 'unknown'`,  since `''` is `falsy`, this silently falls through to  `'unknown'`. The missing-strategy case is handled in two separate places, making it hard to follow

A reader has to hold both lines in their head to understand what the final error message will say. Consider consolidating to `connection?.strategy ?? 'unknown'` directly in the error message and removing the `|| ''` default on line 433, so the fallback is explicit and in one place

src/tools/auth0/handlers/connections.ts
kushalshit27: nit: better to use enums for Management SDK

src/tools/auth0/handlers/connections.ts
ramya18101: The default value is true, But no harm in highlighting!

src/tools/auth0/handlers/connections.ts
ramya18101: Updated

src/tools/auth0/handlers/connections.ts
kushalshit27: its ok to have this

kushalshit27 (approved):

PR #1380, opened 2026-05-07
feat: Add Google Workspace inbound group sync support to directory provisioning

kushalshit27:
### 🔧 Changes

- Add support for `api_enable_groups` on connection options so Google Workspace connections can preserve inbound group provisioning settings during import and export.
- `directory_provisioning_configuration.synchronize_groups` and `synchronized_groups`, which lets Deploy CLI describe whether all groups or only selected groups should sync.
- Update directory provisioning create and update flows to pass `synchronize_groups` through to the Management API, and add synchronized-group replacement logic when the connection uses `synchronize_groups: selected`.

#### YAML format

```yaml
connections:
  - name: google-apps
    strategy: google-apps
    options:
      domain: example.com
      tenant_domain: example.com
      client_id: some_client_id
      client_secret: some_client_secret
      api_enable_groups: true
      api_enable_users: true
    directory_provisioning_configuration:
      synchronize_groups: selected
      synchronized_groups:
        - id: group_123
```

#### JSON format

```json
{
  "name": "google-apps",
  "strategy": "google-apps",
  "options": {
    "domain": "example.com",
    "tenant_domain": "example.com",
    "client_id": "some_client_id",
    "client_secret": "some_client_secret",
    "api_enable_groups": true,
    "api_enable_users": true
  },
  "directory_provisioning_configuration": {
    "synchronize_groups": "selected",
    "synchronized_groups": [
      {
        "id": "group_123"
      }
    ]
  }
}
```

### 🔬 Testing

- Run `npm test` to validate the new schema fields, connection option passthrough, directory provisioning payloads, synchronized-group pagination, and YAML parsing updates.

### 📝 Checklist

### DIFF
--- a/docs/resource-specific-documentation.md
+++ b/docs/resource-specific-documentation.md
@@ -213,6 +213,8 @@ The Deploy CLI supports managing the `directory_provisioning_configuration` for 
 
 The `mapping` array pairs Auth0 user fields with IdP fields, and `synchronize_automatically` controls whether Auth0 runs scheduled sync jobs for the connection.
 
+The `synchronize_groups` field controls group provisioning.
+
 **YAML Example**
 
 ```yaml
@@ -233,6 +235,10 @@ connections:
         - auth0: name
           idp: displayName
       synchronize_automatically: false
+      synchronize_groups: selected
+      synchronized_groups:
+        - id: 'group-id-1'
+        - id: 'group-id-2'
 ```
 
 **Directory Example**
@@ -259,7 +265,9 @@ connections:
       { "auth0": "email", "idp": "mail" },
       { "auth0": "name", "idp": "displayName" }
     ],
-    "synchronize_automatically": false
+    "synchronize_automatically": false,
+    "synchronize_groups": "selected",
+    "synchronized_groups": [{ "id": "group-id-1" }, { "id": "group-id-2" }]
   }
 }
 ```
--- a/docs/resource-specific-documentation.md
+++ b/docs/resource-specific-documentation.md
@@ -224,6 +224,7 @@ connections:
       tenant_domain: example.com
       client_id: 'some_client_id'
       client_secret: 'some_client_secret'
+      api_enable_groups: true
       api_enable_users: true
     directory_provisioning_configuration:
       mapping:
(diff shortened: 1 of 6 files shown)


src/tools/auth0/handlers/connections.ts
```diff
@@ -108,6 +111,20 @@ export const schema = {
             type: 'boolean',
             description: 'The field whether periodic automatic synchronization is enabled',
           },
```
ankita10119, line 113: Just curious, `Management.SynchronizeGroupsEnum` also includes `'off'` as a valid value. Should we add a brief note in the docs or a comment here about what `'off' `
means in practice? For example, is it the same as omitting the field, or does it explicitly disable group sync? I wasn't sure if users would need to know when to
use `'off'` vs just removing the field.

test/tools/auth0/handlers/connections.tests.js
ankita10119, line 345: Should we also extend this existing getType test (or add one alongside it) to cover the new `synchronize_groups / synchronized_groups` fields on the export path? I was thinking when `dirProvConfigs` returns an entry with `synchronize_groups: 'selected'`, should we verify that `getConnectionSynchronizedGroups` is called and its result shows up as `synchronized_groups` on the exported connection? I may be missing something, but since this is what users see in their `tenant.yaml` after a0deploy export, I wanted to check if we need a test for that path too

docs/resource-specific-documentation.md
ankita10119, line 214: Should we update this section to also describe the new `synchronize_groups` and `synchronized_groups` fields? The current prose and both code examples only show
mapping and synchronize_automatically. The PR description has really good YAML/JSON examples for the new fields , would it make sense to bring those into the docs file as well so users can find them here?

src/tools/auth0/handlers/connections.ts
kushalshit27: The detailed doc should be on the Auth0 doc  and Management API, I can add a one-liner here

docs/resource-specific-documentation.md
kushalshit27: Sure

kushalshit27: Updated

test/tools/auth0/handlers/connections.tests.js
kushalshit27: Updated

ankita10119 (approved):

PR #1383, opened 2026-05-12
feat: add organization_id support to connected_accounts schema for Token Vault org binding

harshithRai:
### 🔧 Changes

* Added `organization_id` as an optional string property to the `connected_accounts` schema in the connections handler. 
   * This prevents schema validation from rejecting or stripping the `organization_id` field returned by the Management API (`GET /v2/users/{id}/connected-accounts`) for accounts linked under an organization.

### 🔬 Testing

* Added a unit test verifying that a connection with `connected_accounts: { active: true, organization_id: 'org_abc123' }` correctly flows through `processChanges` and reaches the API update payload without being stripped or rejected.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/connections.ts
+++ b/src/tools/auth0/handlers/connections.ts
@@ -90,7 +90,7 @@ export const schema = {
         type: 'object',
         properties: {
           active: { type: 'boolean' },
-          organization_id: { type: 'string' },
+          organization_name: { type: 'string' },
         },
         required: ['active'],
         additionalProperties: false,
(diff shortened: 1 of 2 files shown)


src/tools/auth0/handlers/connections.ts
```diff
@@ -90,6 +90,7 @@ export const schema = {
         type: 'object',
         properties: {
           active: { type: 'boolean' },
+          organization_id: { type: 'string' },
```
kushalshit27: - Can we update the `organization_id` ? Is it supported by SDK ?
- `organization_id `should be organization `name` while exporting, we generally do not expert id by default, name base linking happens.

kushalshit27 (commented):
Please add proper PR description , check old feature request Prs

kushalshit27 (commented):
Please make sure your commits are singed to make it merge ready

kushalshit27 (changes requested):

harshithRai:
Closing this PR after further investigation.

The organization_id field added in the `management-api` exists on the user-level `ConnectedAccount` response (`GET /v2/users/{id}/connected-accounts`), not on the connection configuration object. 

The Deploy CLI manages tenant configuration, not runtime user data, so there is no change needed here.

PR #1385, opened 2026-05-13
feat: configurable ignoreDryRunFields via AUTH0_IGNORE_DRY_RUN_FIELDS

dotjoshrc:
### 🔧 Changes

Adds `AUTH0_IGNORE_DRY_RUN_FIELDS`, a new optional config key mapping handler type to an array of field paths to exclude from `--dry-run` diff comparisons. User-supplied entries are merged additively with each handler's built-in `ignoreDryRunFields` defaults, so the curated defaults (`smtp.credentials.smtp_pass`, `mandrill.credentials.api_key`, `_clientName`, etc.) cannot be accidentally dropped.

**Motivation.** The Management API does not echo secret values on read — `client_secret` on clients, `options.client_secret` on connections, action `secrets[].value`, and the email provider's `credentials.api_key` for non-SMTP/Mandrill providers (notably Mailgun). The corresponding local YAML fields therefore always surface as diffs in `--dry-run` output, drowning out signal on every run. Today the only knob is per-handler hardcoded arrays; the goal here is to let tenant configs silence the known-noisy fields for their setup without forking the CLI.

**Shape.** New optional config key, identical in shape to `EXCLUDED_PROPS`/`INCLUDED_PROPS`:

```jsonc
{
  "AUTH0_IGNORE_DRY_RUN_FIELDS": {
    "clients": ["client_secret"],
    "connections": ["options.client_secret"],
    "actions": ["secrets"],
    "emailProvider": ["credentials.api_key"]
  }
}
```

This only affects `--dry-run` reporting; a real import still sends every local field to the API as before.

**Backwards compatibility.** When the key is absent the handler behavior is byte-identical to today. The constructor read is wrapped in a `try/catch` so handlers constructed against an uninitialized `configFactory()` (as some unit tests do) continue to work.

**Files touched.** `src/types.ts`, `src/tools/auth0/handlers/default.ts`, `test/tools/auth0/handlers/default.tests.ts`, `docs/configuring-the-deploy-cli.md`, `CHANGELOG.md`. 119 LOC.

### 📚 References

N/A — no linked issue. Validated against a real tenant (Mailgun email provider, OIDC connection with `client_secret`, action with five `secrets[]` entries) where it cleanly suppresses the previously-noisy diff lines.

### 🔬 Testing

- New unit tests in `test/tools/auth0/handlers/default.tests.ts` covering: (1) merge of user-configured fields with handler defaults, (2) dedupe of overlapping entries, (3) isolation across handler types, (4) fallback when the config provider returns undefined.
- Full suite: `npm test` → 1228 passing / 1 pending (unchanged from main).
- `npm run lint`, `npm run format:check`, `npm run build`, `npx kacl lint` all clean.

Manual verification: ran `a0deploy import --dry-run --debug -c <stage.json> -i tenant.yaml` against a real Auth0 tenant; before this change, every dry-run logged a `found in 'localObj' but not in 'remoteObj'` line for `mailgun.credentials.api_key` and each entry of `handle-post-login.secrets`. With the config set, those become `Ignoring key … due to ignoreDryRunFields configuration` debug lines instead, and the noise disappears at non-debug levels.

### 📝 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `AUTH0_IGNORE_DRY_RUN_FIELDS` configuration option, a map of handler type → field paths to exclude from `--dry-run` diff comparisons. Merges additively with each handler's built-in defaults so users can suppress noise from fields the Management API never returns (e.g. `client_secret`, action `secrets`, email provider `credentials.api_key`) without losing the curated defaults.
+
 ## [8.34.0] - 2026-05-11
 
 ### Added
--- a/docs/configuring-the-deploy-cli.md
+++ b/docs/configuring-the-deploy-cli.md
@@ -191,6 +191,25 @@ Provides ability to exclude any unwanted properties from management.
 }
 ```
 
+### `AUTH0_IGNORE_DRY_RUN_FIELDS`
+
+Object keyed by resource handler type, with values that are arrays of field paths to exclude from `--dry-run` diff comparisons. Useful for suppressing noisy diffs on fields the Management API never returns (e.g. secret values), so dry-run output focuses on changes you can actually verify.
+
+User-supplied fields are merged with the handler's built-in defaults; they do not replace them. Field paths support exact match, `.endsWith('.<field>')` suffix match, and `.endsWith('[<field>]')` suffix match (see `calculateDryRunChanges#shouldIgnoreDryRunField`).
+
+This option only affects `--dry-run` reporting. A non-dry-run import still sends every local field to the Management API.
+
+#### Example
+
+```json
+{
+  "clients": ["client_secret"],
+  "connections": ["options.client_secret"],
+  "actions": ["secrets"],
+  "emailProvider": ["credentials.api_key"]
+}
+```
+
 ### `AUTH0_AUDIENCE`
 
 String. Separate value from audience value while retrieving an access token for management API. Useful when default Management API endpoints are not publicly exposed.
--- a/src/tools/auth0/handlers/default.ts
+++ b/src/tools/auth0/handlers/default.ts
@@ -171,6 +171,29 @@ export default class APIHandler {
     this.deleted = 0;
   }
 
+  /**
+   * Returns the effective `ignoreDryRunFields` for this handler, merging the
+   * handler's hardcoded defaults with any user-supplied entries from the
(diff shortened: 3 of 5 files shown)


dotjoshrc:
The `E2E tests as Node module` check is failing, but the failure is unrelated to this PR. Surfacing the analysis here so reviewers don't have to re-derive it:

**Failure pattern** (4 tests, identical across re-runs):
- `#end-to-end deploy: should deploy without throwing an error`
- `#end-to-end deploy: should deploy without deleting resources if AUTH0_ALLOW_DELETE is false`
- `#end-to-end deploy: should deploy while deleting resources if AUTH0_ALLOW_DELETE is true`
- `#end-to-end dump and deploy cycle: should dump and deploy without throwing an error`

All four fail with the same nock mismatch — `PATCH` / `DELETE` on the **Deploy CLI** client (`client_id: Vp0gMRF8PtMzekil38qWoj4Fjw2VjRZE`, the client used to authenticate the test itself). That client is supposed to be filtered out of changesets by `clients.ts:589`:

```ts
item.client_id !== currentClient
```

…where `currentClient = this.config('AUTH0_CLIENT_ID')`. The nock recordings (`test/e2e/recordings/should-deploy-without-throwing-an-error.json` and siblings) contain zero requests against `Vp0gMRF8…`, confirming the test was recorded with the auth-client filter active.

**Why this isn't from this PR:**

1. `src/tools/auth0/handlers/default.ts` constructor is **byte-identical to upstream master** in the current revision of this branch — `git diff upstream/master -- src/tools/auth0/handlers/default.ts` shows only an added `getEffectiveIgnoreDryRunFields()` method plus two call-site swaps inside the existing `if (AUTH0_DRY_RUN)` and `dryRunChanges` branches.
2. The four failing tests invoke `deploy()` with configs that contain only `AUTH0_DOMAIN`/`AUTH0_CLIENT_ID`/`AUTH0_CLIENT_SECRET`/`AUTH0_ACCESS_TOKEN` — no `AUTH0_DRY_RUN`. So `getEffectiveIgnoreDryRunFields()` is never called on any code path exercised by these tests.
3. The other e2e tests in the same suite that do exercise `deploy()` but with `AUTH0_INCLUDED_ONLY: ['tenant']` (skipping the clients handler entirely) all pass.

**What I'd suggest:** treat the four failing tests as a pre-existing CI flake on the auth-client filter and re-run, or point me at the right place to look if there's something more I can do from this PR. Happy to debug further if you'd like.

kushalshit27:
Thank you for submitting this PR! Your contribution is greatly appreciated. We'll review it shortly

kushalshit27 (approved):
LGTM

dotjoshrc:
Totally missed the approval and merge 🎉 Much appreciated.

PR #1387, opened 2026-05-15
fix: preserve keyword markers in themes for default export configuration

ankita10119:
### 🔧 Changes

 Fixes keyword preservation for themes when `AUTH0_EXPORT_IDENTIFIERS` is `false` (the default configuration).
                                                                                                                                                                               
  **Root cause:** `ThemesHandler` used `identifiers: ['themeId']` to match local theme entries during keyword preservation. However, `themeId` is stripped from exported files by default (`AUTH0_EXPORT_IDENTIFIERS: false`), so the preservation logic could never find a matching identifier and silently skipped preservation, causing keyword markers 
  like `##CDN_URL##` to be overwritten with raw values on re-export.                                                                                                         
                                         
  **Changes:**
  - `ThemesHandler`: Added `displayName` as a secondary identifier alongside `themeId`, providing a match path for users who have named their theme but do not use `AUTH0_EXPORT_IDENTIFIERS`
  - `getPreservableFieldsFromAssets` in `keywordPreservation.ts`:
    - Fixed the identifier `reduce` to skip missing identifiers instead of resetting the accumulator, allowing multi-identifier fallback to work correctly
    - Added positional index fallback when registered identifiers are absent from the local file, safe for themes since Auth0 supports only one theme per tenant

### 📚 References

 - Related: #1379 (partial fix - only covered `AUTH0_EXPORT_IDENTIFIERS: true`)

### 🔬 Testing

Unit tests added covering the following scenarios:

  1. `themeId` present in local file (`AUTH0_EXPORT_IDENTIFIERS: true`) - preserved via `themeId` identifier
  2. `displayName` present but not `themeId` - preserved via `displayName` identifier
  3. Neither `themeId` nor `displayName` present (default export, exact customer scenario) - preserved via positional index fallback
  4. `identifiers: []` (no identifiers registered for array type) - NOT preserved, confirming the fallback does not apply to unrecognized arrays

  **Manual end-to-end test:**
  1. Set `AUTH0_PRESERVE_KEYWORDS: false` in config, run export to get a baseline `tenant.yaml`
  2. Edit `tenant.yaml` - replace raw URLs with quoted keyword markers e.g. `font_url: '##CDN_URL##/fonts/custom.woff2'`
  3. Set `AUTH0_PRESERVE_KEYWORDS: true`, re-run export
  4. Confirm `tenant.yaml` still contains `##CDN_URL##` markers instead of resolved URLs

### 📝 Checklist

### DIFF
--- a/src/keywordPreservation.ts
+++ b/src/keywordPreservation.ts
@@ -81,10 +81,7 @@ export const getPreservableFieldsFromAssets = (
         );
 
         if (specificAddress.length === 0) {
-          // If identifiers are registered for this resource type but none were found on the
-          // local array item (e.g. themeId stripped by default export), fall back to positional
-          // index so keyword markers are still preserved. Safe for singletons like themes.
-          // If no identifiers are registered at all (unrecognized nested arrays), skip.
+          // No identifiers registered: skip. Identifiers registered but absent from item: fall back to positional index.
           if (resourceIdentifiers.length === 0) {
             return [];
           }
--- a/src/keywordPreservation.ts
+++ b/src/keywordPreservation.ts
@@ -66,10 +66,10 @@ export const getPreservableFieldsFromAssets = (
         const specificAddress = resourceIdentifiers.reduce(
           (aggregateAddress, resourceIdentifier) => {
             resourceSpecificIdentifiers[address];
-            if (resourceIdentifier === undefined) return ''; // See if this specific resource type has an identifier
+            if (resourceIdentifier === undefined) return aggregateAddress; // See if this specific resource type has an identifier
 
             const identifierFieldValue = arrayItem[resourceIdentifier];
-            if (identifierFieldValue === undefined) return ''; // See if this specific array item possess the resource-specific identifier
+            if (identifierFieldValue === undefined) return aggregateAddress; // See if this specific array item possess the resource-specific identifier
 
             if (aggregateAddress === '') {
               return `${resourceIdentifier}=${identifierFieldValue}`;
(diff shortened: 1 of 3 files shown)


src/tools/auth0/handlers/themes.ts
```diff
@@ -429,7 +429,7 @@ export default class ThemesHandler extends DefaultHandler {
       ...options,
       type: 'themes',
       id: 'themeId',
-      identifiers: ['themeId'],
+      identifiers: ['themeId', 'displayName'],
```
kushalshit27: The displayName fallback only works correctly when local and remote displayName are in sync.

kushalshit27 (dismissed):
LGTM

kushalshit27 (dismissed):

src/keywordPreservation.ts
```diff
@@ -81,7 +81,20 @@ export const getPreservableFieldsFromAssets = (
         );
 
         if (specificAddress.length === 0) {
-          return [];
+          // If identifiers are registered for this resource type but none were found on the
+          // local array item (e.g. themeId stripped by default export), fall back to positional
+          // index so keyword markers are still preserved. Safe for singletons like themes.
+          // If no identifiers are registered at all (unrecognized nested arrays), skip.
```
kushalshit27: Suggestion: keep the comment generic and as short as possible.

ankita10119: Addressed and updated

kushalshit27:
Please check the failing tests

kushalshit27 (approved):

PR #1388, opened 2026-05-15
fix: prevent path traversal in config file handlers

ankita10119:
### 🔧 Changes

 Fixes a path traversal vulnerability in config file handlers where user-supplied file paths in configuration (e.g. `code`, `customScripts`) could reference files outside the config directory using sequences like `../../`.                                                                                                                             
                                         
  **Changes:**                                                                                                                                                                 
  - `DirectoryContext.loadFile`: Resolves the final path and validates it stays within the config root before loading. Removes the previous fallback that silently tried the raw path if the relative path didn't exist, that fallback was the entry point for traversal                                                                                 
  - `yaml/index.ts`: Same boundary check applied to the YAML context's `loadFile`, throwing an explicit error if the resolved path escapes the config directory
  - `databases.ts`: Added config root boundary check for `customScripts` file paths defined in `database.json`
  - `actions.ts` / `actionModules.ts`: Removed the `fs.existsSync` branch that allowed absolute or root-relative paths to bypass the relative-path resolution, paths are now always resolved relative to the handler folder

### 🔬 Testing

 - Updated existing action tests to use relative paths (`./actions/code.js`) instead of absolute paths, reflecting the new enforcement that file references must stay within the config directory
  - Updated YAML pages test to use relative `page.html` instead of an absolute path, and fixed file write ordering to match the new resolution logic
  - Path traversal attempts (e.g. `../../etc/passwd` as a script path) will now throw an explicit error rather than silently loading the file

### 📝 Checklist


kushalshit27 (commented):
This could be a breaking change — absolute and CWD-relative paths are no longer accepted
 
The removed fallbacks (toLoad = f in YAML context, fs.existsSync(unixPath) branch in actions) were also the mechanism that allowed users to specify absolute paths or paths relative to the working directory. Users with existing configs like this will now get a hard error: 

 ### YAML — was valid before, now throws  
```                                                                                          
  pages:                                                                                                                           
    - name: login                                                                                                                  
      html: "/home/ci/scripts/login.html"      
```                                                                                    
                                                                                                                                   
 ### action.json — was valid before, now throws       
```                                                                             
  { "code": "./local/scripts/action.js" } 
```

ankita10119:
> This could be a breaking change — absolute and CWD-relative paths are no longer accepted
> 
> The removed fallbacks (toLoad = f in YAML context, fs.existsSync(unixPath) branch in actions) were also the mechanism that allowed users to specify absolute paths or paths relative to the working directory. Users with existing configs like this will now get a hard error:
> 
> ### YAML — was valid before, now throws
> ```
>   pages:                                                                                                                           
>     - name: login                                                                                                                  
>       html: "/home/ci/scripts/login.html"      
> ```
> 
> ### action.json — was valid before, now throws
> ```
>   { "code": "./local/scripts/action.js" } 
> ```

You are right @kushalshit27  
The removed fallbacks were indeed the mechanism for absolute and CWD relative paths, so this is a breaking change for users relying on those patterns. Since the goal here is a security fix, keeping support for absolute paths would conflict with the boundary enforcement

Updated the error messages to guide users toward config-relative paths instead of surfacing a "path traversal" warning, which could be confusing for legitimate use cases. Will add a CHANGELOG entry to call this out explicitly.

kushalshit27:
@ankita10119 ,
Is it possible to make it backwards-compatible while fixing the issue for `exports`?

test/context/directory/actions.test.js
```diff
@@ -15,7 +15,7 @@ const actionFiles = {
       '/** @type {PostLoginAction} */ module.exports = async (event, context) => { console.log(@@replace@@); return {}; };',
     'action-one.json': `{
       "name": "action-one",
-      "code": "./local/testData/directory/test1/actions/code.js",
```
kushalshit27, line 18: 2 set Unit test should be add:
- old path
- new path

kushalshit27 (changes requested):

test/context/directory/actions.test.js
ankita10119: Addressed and updated

ankita10119:
> @ankita10119 , Is it possible to make it backwards-compatible while fixing the issue for `exports`?

I understand the concern @kushalshit27 

But making this backwards-compatible would conflict with actually fixing the security issue. Any approach that still falls through to the raw path (absolute or CWD-relative) for unrecognized references would allow a config file to reference files like `/etc/passwd` or `../../secrets/.env`, which would then be read and uploaded to Auth0.                                                                                                                                                
                                                                                                                                                                               
The realistic attack vector is a CI pipeline pulling config from a compromised source where a malicious path silently exfiltrates sensitive files into Auth0. The hard error is necessary to close that                                                                                                                                              

I've added explicit test coverage for both the rejection case (path outside config root) and the happy path (relative path within config dir), and will document this as a breaking change in the CHANGELOG.

PR #1389, opened 2026-05-18
fix: remove stale connection files from directory export when connections are deleted

### LINKED ISSUE #1376: Social Connections not deleting on import with AUTH0_ALLOW_DELETE=true
### Checklist

### Description

A source/export tenant with zero social connections present does not remove social connections that are present in downstream destination/import tenants.

This seems like either a bug with how the deploy tool handles an empty directory, or I'm not understanding how the AUTH0_ALLOW_DELETE functions.

### Expectation

The expectation is that it would delete any social connection present in the stage/destination tenant.

### Reproduction

1. In the dev/source tenant, delete all social connections, leaving none present.
2. Export the tenant config to code (using directory if that matters). This creates an empty 'connections' folder, which is expected.
3. Configure `AUTH0_ALLOW_DELETE` to `true` in the destination config.json file
4. Deploy an import action to the destination tenant
5. Any social connection present in that tenant remains and is not deleted.

### Deploy CLI version

8.32.0

### Node version

24.15.0

harshithRai:
Fixes a stale file issue in the directory format export where connection JSON files left over from previous exports were not removed when connections were deleted from the source tenant, causing subsequent imports to treat them as desired state and skip deletion from destination tenants.

### 🔧 Changes

When re-exporting a tenant after social connections have been removed, the `dump()` function in the directory connections handler now removes JSON files that no longer correspond to a current connection. Previously, stale files from prior exports remained in the `connections/` folder, causing subsequent imports to treat those connections as desired state and never trigger their deletion from destination tenants.

* `src/context/directory/handlers/connections.ts` — updated `dump()` to remove stale `.json` and `.html` files from the `connections/` directory that no longer correspond to a connection in the current export
* `test/context/directory/connections.test.js` — added tests covering empty connections export, partial removal of stale files, and full removal of stale files

### 📚 References

Fixes #1376

### 🔬 Testing

Added two new tests to validate that:

* Re-exporting with a subset of connections removes the stale JSON files for deleted connections
* Re-exporting with zero connections removes all stale JSON files and the empty folder parses back as `[]`, enabling import to trigger deletions

### 📝 Checklist

### DIFF
--- a/src/context/directory/handlers/connections.ts
+++ b/src/context/directory/handlers/connections.ts
@@ -76,29 +76,7 @@ async function dump(context: DirectoryContext): Promise<void> {
   }
 
   const connectionsFolder = path.join(context.filePath, constants.CONNECTIONS_DIRECTORY);
-  fs.ensureDirSync(connectionsFolder);
-
-  // Build the set of expected filenames for the current connections so stale
-  // files left over from previously-exported connections can be removed.
-  const expectedFiles = new Set<string>(
-    connections.flatMap((connection) => {
-      const connectionName = sanitize(connection.name);
-      const files = [`${connectionName}.json`];
-      if (connection.strategy === 'email') {
-        files.push(`${connectionName}.html`);
-      }
-      return files;
-    })
-  );
-
-  // Remove stale files that no longer correspond to a current connection.
-  getFiles(connectionsFolder, ['.json', '.html']).forEach((filePath) => {
-    const fileName = path.basename(filePath);
-    if (!expectedFiles.has(fileName)) {
-      log.info(`Deleting stale connection file: ${filePath}`);
-      fs.removeSync(filePath);
-    }
-  });
+  fs.emptyDirSync(connectionsFolder);
 
   // Convert enabled_clients from id to name
   connections.forEach((connection) => {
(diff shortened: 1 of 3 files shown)


src/context/directory/handlers/connections.ts
```diff
@@ -76,7 +76,7 @@ async function dump(context: DirectoryContext): Promise<void> {
   }
 
   const connectionsFolder = path.join(context.filePath, constants.CONNECTIONS_DIRECTORY);
-  fs.ensureDirSync(connectionsFolder);
+  fs.emptyDirSync(connectionsFolder);
```
ankita10119: Could this cause an unintended regression with excluded connections? If a user has a pre-existing connections/ folder from a prior export and re-runs dump with some         
connections listed under exclude, emptyDirSync would wipe those excluded connections' files before the exclusion filter runs, meaning files for excluded (but not deleted) connections would be lost from the local directory. With `ensureDirSync` those files were preserved.

Would it be worth adding a test that pre-populates the folder, sets an exclusion, and verifies the excluded connection's file still exists after dump? Curious if the intent is for exclude to also remove files from local, or just skip writing them

ankita10119 (changes requested):

src/context/directory/handlers/connections.ts
harshithRai: That's a valid catch. It does delete when a connection is exported first and then added to the exclusion list, when ideally it should be excluded. I'll revise the fix approach accordingly and push an update.

harshithRai: Fixed it so instead of clearing the folder upfront, we write all the connections first and then only delete files that don't belong to an active or excluded connection.

ankita10119 (approved):

PR #1401, opened 2026-06-12
fix: remove enable_custom_domain_in_emails from managed tenant flags

ankita10119:
### 🔧 Changes

Removes `enable_custom_domain_in_emails` from the managed tenant flags allowlist (`allowedTenantFlags`) to prevent import failures in environments without a ready custom domain.

  **Problem:** The flag was included in `allowedTenantFlags`, causing it to be:
  1. Included in `export` output (`tenant.yaml`) even though it is deprecated
  2. Sent to the Management API during `import`, resulting in a `400 Bad Request` error on tenants that do not have a custom domain with `status: "ready"` - aborting the entire tenant `processChanges` stage and leaving no tenant settings applied

  **Changes:**
  - Removed `enable_custom_domain_in_emails` from `allowedTenantFlags` in `src/tools/auth0/handlers/tenant.ts` - the existing `removeUnallowedTenantFlags()` function now strips it on both the export and import paths automatically
  - Updated the deprecation warning message to accurately state the flag is no longer applied, and directs users to migrate to `customDomains[].is_default`
  - Updated unit test to assert the flag is absent from the Management API payload, not just that the warning is logged
  - Removed the flag from three e2e testdata fixtures that used it as import input

  **Migration:** Users should remove `enable_custom_domain_in_emails` from their tenant configuration files and use the `is_default` field on `customDomains` instead:

  ```yaml
  customDomains:
    - domain: mail.example.com
      type: auth0_managed_certs
      tls_policy: recommended
      is_default: true
```

### 📚 References

 - Supersedes the deprecation warning introduced in #1330 -the flag is now fully removed from management rather than just warned about

### 🔬 Testing

Unit tests: Updated test in `test/tools/auth0/handlers/tenant.tests.ts` verifies:
  - Deprecation warning is still logged when the flag is present in input
  - The flag is not present in the payload sent to client.tenants.settings.update()

  Manual export test: Running node lib/index.js export against a tenant with `enable_custom_domain_in_emails: true `confirms:
  - The flag no longer appears in the exported tenant.yaml
  - The export log lists the flag under the existing "incompatible flags" warning

  Not tested end-to-end for import (requires a tenant without a custom domain): the import path is covered by the unit test and by code analysis - `removeUnallowedTenantFlags()` strips the flag before the API call in processChanges().

### 📝 Checklist

### CODE REVIEW

File: src/tools/auth0/handlers/tenant.ts [hunk 1]
```diff
@@ -233,9 +232,9 @@ export default class TenantHandler extends DefaultHandler {
 
     if ((tenant.flags as Record<string, unknown>)?.enable_custom_domain_in_emails !== undefined) {
       log.warn(
-        'The "enable_custom_domain_in_emails" tenant flag is deprecated. ' +
-          'Use the "is_default" field on customDomains to configure the default domain instead. ' +
-          'The flag will still be applied for now but will be removed in a future release.'
+        'The "enable_custom_domain_in_emails" tenant flag is deprecated and has been removed from management. ' +
+          'It will not be applied during import. ' +
+          'Use the "is_default" field on customDomains to configure the default domain instead.'
       );
```
harshithRai: Now that `enable_custom_domain_in_emails` is removed from `allowedTenantFlags`, `removeUnallowedTenantFlags()` will already emit a generic `incompatible flag` warning for it. 

So in the same import run the user sees two contradictory warnings, one says the flag _can likely be removed_ from their config, the other says it has _already been removed_ from management.

For example, with flags: `{ enable_custom_domain_in_emails: true }` in their config, they'd see:

```bash
WARN: The following tenant flag has not been updated because deemed incompatible with the target tenant: enable_custom_domain_in_emails. This flag can likely be removed from the tenant definition file...

WARN: The "enable_custom_domain_in_emails" tenant flag is deprecated and has been removed from management. It will not be applied during import. Use the "is_default" field on customDomains instead.
```

ankita10119: Addressed and updated

harshithRai (approved):

PR #1407, opened 2026-06-23
feat: add fedcm_login support for Google One Tap / FedCM in Universal Login

ankita10119:
### 🔧 Changes

Adds fedcm_login to the client JSON schema to support the Google One Tap / FedCM (Federated Credential Management) feature in New Universal Login.                           
                                                                                                                                                                               
  Schema addition (`src/tools/auth0/handlers/clients.ts`):
  - fedcm_login - optional object (null supported for deletion via PATCH)                                                                                                      
    - fedcm_login.google - required if fedcm_login is defined                                                                                                                  
        - fedcm_login.google.is_enabled - boolean, controls whether the Google One Tap prompt is shown on New Universal Login for this client                                  

  No additional handler changes were required - fedcm_login is a plain nested object with no ID-to-name mapping, no special serialization, and is valid on both create and
  update operations.

### 📚 References



### 🔬 Testing

 Unit tests (`test/tools/auth0/handlers/clients.tests.js`):
  - should pass validation with fedcm_login - { google: { is_enabled: true } } accepted by schema
  - should pass validation with fedcm_login set to null - null accepted by schema (deletion case)
  - should create client with fedcm_login - fedcm_login passed through unchanged to the API create call

Manual end-to-end against dev-ankita-t.us.auth0.com (feature flag enable_google_one_tap_ul enabled):

| Scenario | Operation | Result |
|--------|--------|--------|
| is_enabled: true on existing client | PATCH | Written and returned in export |
| is_enabled: false on existing client | PATCH | Written and returned in export |
| fedcm_login: null   | PATCH | Field absent from export |
| New client with fedcm_login  | POST (create) | Written and returned in export | 


 
### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/clients.ts
+++ b/src/tools/auth0/handlers/clients.ts
@@ -403,6 +403,18 @@ export const schema = {
           },
         },
       },
+      fedcm_login: {
+        type: ['object', 'null'],
+        description: 'Configuration for FedCM (Federated Credential Management) login',
+        properties: {
+          google: {
+            type: 'object',
+            properties: {
+              is_enabled: { type: 'boolean' },
+            },
+          },
+        },
+      },
     },
     required: ['name'],
   },
--- a/src/tools/auth0/handlers/clients.ts
+++ b/src/tools/auth0/handlers/clients.ts
@@ -412,8 +412,10 @@ export const schema = {
             properties: {
               is_enabled: { type: 'boolean' },
             },
+            required: ['is_enabled'],
           },
         },
+        required: ['google'],
       },
     },
     required: ['name'],
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "dependencies": {
     "@clack/prompts": "1.3.0",
     "ajv": "^6.12.6",
-    "auth0": "^5.12.0",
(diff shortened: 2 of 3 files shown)


src/tools/auth0/handlers/clients.ts
```diff
@@ -403,6 +403,18 @@ export const schema = {
...
+        description: 'Configuration for FedCM (Federated Credential Management) login',
+        properties: {
+          google: {
+            type: 'object',
+            properties: {
+              is_enabled: { type: 'boolean' },
+            },
+          },
+        },
+      },
```
harshithRai, line 419: As per the description, `google` is required when `fedcm_login` is set. But the schema doesn't enforce this; `fedcm_login: {}` and `fedcm_login: { google: {} }` both pass validation based on current code. 

Similar fields like `token_quota` handle this with `required: ['client_credentials']`.

Maybe we should add `required: ['google']` under `fedcm_login` and `required: ['is_enabled']` under `google`?

harshithRai (approved):

src/tools/auth0/handlers/clients.ts
ankita10119: Addressed and updated

PR #1415, opened 2026-06-30
Stop export errors from being swallowed

joshcanhelp:
I was trying to export an Auth0 entity and my M2M client did not have the correct permissions. The error was a generic "Error: Problem exporting clientGrants"

### 🔧 Changes

Minor update to provide the error encountered when exporting entities.

### 🔬 Testing

Export an entity without having the correct permissions before and after the change to see the difference in output.

### 📝 Checklist

- **N/A** I have added documentation for all new/changed functionality


kushalshit27:
Thank you for submitting this PR!  We'll review it shortly.

kushalshit27:
All commits should be signed. Please update the commit to make the PR merge-ready. Thanks

ankita10119:
Hi @joshcanhelp 

You added a second argument (`{ cause }`) to the Error constructor, which is not supported in ES2020, this is causing the CI failures. Since `tsconfig.json` has `"target": "es2020"`, TypeScript raises `TS2554: Expected 0-1 arguments, but got 2`, which also causes the E2E check to fail downstream. Could you update `tsconfig.json` to `"target":"es2022"` to fix this? It's safe since `node >= 20.19.0` is already required                                                                                            
                                                                                                                                                                            
Also, please make sure to address @kushalshit27 comment as well

joshcanhelp:
Sorry for the delay, I was OOO. I'll take a look this week!

joshcanhelp:
@kushalshit27 @ankita10119 - I made the tsconfig change in a signed commit so I think we're good here!

joshcanhelp:
Looks like the target change might have generated some TS issues. The E2E tests don't look related to this change and I'm not sure about the unit test timeout.

PR #1417, opened 2026-07-02
feat: add client credential lifecycle management for Private Key JWT and mTLS

### LINKED ISSUE #1167: Ability to create a client credential accepting a public key
### Checklist

### Describe the problem you'd like to have solved

Hi team, we're interested in using the client credential feature [Private Key JWT Authentication](https://auth0.com/docs/secure/application-credentials#private-key-jwt-authentication). 

From what I can tell, the Auth0 Deploy CLI allows setting a client credential for an application by pointing at its key ID. 

```
  "client_authentication_methods": {
    "private_key_jwt": {
      "credentials": [
        {
          "id": "cred_12345"
        }
      ]
    }
  },
```

But it does not allow actually passing those public keys so we can create credential IDs. 

Feature request is to allow us to also specify the public keys here so that we don't really have to pass a specific credential ID. 

### Describe the ideal solution

Ideal solution - we pass up to 2 maximum public keys strings for a client application, and Auth0 Deploy CLI manages creating and managing it. When we remove a key and replace it with another, Auth0 Deploy CLI manages removing the old one and putting the new one in place. This is a nice way to rotate client credentials. 

### Alternatives and current workarounds

I'm not sure what to do

### Additional context

_No response_

harshithRai:
### 🔧 Changes

The CLI previously had no support for the `/clients/{id}/credentials` API. This adds full create/delete reconciliation for client credentials across all three credential types (`public_key`, `x509_cert`, `cert_subject_dn`).

**Export changes**

- `client_authentication_methods` is now exported for **all** client types (previously only `express_configuration` clients). This fixes incomplete export behavior for M2M and other app types.
- Credential stubs are enriched with full metadata (`name`, `credential_type`, `kid`, `alg`) instead of just `{ "id": "cred_xxx" }`.
- `pem` is never exported; Auth0 does not return it after creation.

Before:
```json
"client_authentication_methods": {
  "private_key_jwt": {
    "credentials": [{ "id": "cred_abc123" }]
  }
}
```

After:

```json
"client_authentication_methods": {
  "private_key_jwt": {
    "credentials": [{
      "id": "cred_abc123",
      "name": "my-rotation-key",
      "credential_type": "public_key",
      "kid": "9P8IC1...",
      "alg": "RS256"
    }]
  }
}
```

**Deploy changes**

A new clientAuthCredentials handler (runs at order 70, after clients at 50) reconciles credential state when the config contains credentials with a pem field:

User-managed config shape (pem must be supplied manually from key generation; Auth0 never returns it):

```json
"client_authentication_methods": {
  "private_key_jwt": {
    "credentials": [{
      "name": "my-rotation-key",
      "pem": "-----BEGIN PUBLIC KEY-----\n...",
      "credential_type": "public_key"
    }]
  }
}
```

* Creates credentials present in config but missing in Auth0
* Deletes credentials removed from config (requires `AUTH0_ALLOW_DELETE=true`)
* Creates run before deletes; Auth0 allows max 2 credentials per client; this preserves the rotation overlap window and avoids an auth gap
* Re-wires `client_authentication_methods` on the client with the resolved credential IDs after reconciliation

**Safety / backward compatibility**

* Credentials are only managed when the config contains a pem field. Configs with no credentials, id-only credentials `{ "id": "cred_xxx" }`, or no `client_authentication_methods` at all are untouched, ensuring there is no breaking change for existing users.
* Deletions are gated behind `AUTH0_ALLOW_DELETE`, consistent with other resource types.
* Creates always run before deletes, so both credentials exist simultaneously during the rotation window - no auth gap.
* A plain `export`->`deploy` never deletes credentials; reconciliation only activates when at least one credential in the config has a `pem` field

### 📚 References

Closes #1167

### 🔬 Testing

Manual end-to-end testing against a dev tenant:

1. **Export enrichment**: registered a credential via the API directly, ran `auth0 export`, verified exported JSON includes full metadata (`name`, `kid`, `alg`, `credential_type`) for all client types, not just `express_configuration`
2. **Create**: added a new credential with `pem` to the config, ran `auth0 deploy`, verified it was created in Auth0
3. **Rotation overlap**: added a second credential alongside an existing one, verified both existed simultaneously in Auth0 before the old one was removed
4. **`AUTH0_ALLOW_DELETE=false` guard**: removed a credential from config with `AUTH0_ALLOW_DELETE=false`, verified a warning was logged and Auth0 state was untouched
5. **Delete**: with `AUTH0_ALLOW_DELETE=true`, removed a credential from config, verified it was deleted from Auth0 after `client_authentication_methods` was updated
6. **Full cleanup**: removed a pem-based credential from config entirely, verified it was deleted from Auth0 after `client_authentication_methods` was updated
7. **Idempotency**: deployed with no config changes, verified no create/delete operations ran
8. **Safety gate**: deployed config with no `pem` fields, verified credentials in Auth0 were untouched

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/clientAuthCredentials.ts
+++ b/src/tools/auth0/handlers/clientAuthCredentials.ts
@@ -98,8 +98,10 @@ export default class ClientAuthCredentialsHandler {
         continue;
       }
 
-      // If no pem-based credentials in config and nothing in Auth0, nothing to do
-      if (desired.length === 0 && existing.length === 0) continue;
+      // Skip reconciliation entirely when no pem is present — pem is the opt-in signal.
+      // Without this guard, a plain export→deploy would delete all existing credentials
+      // because desired would be empty (exported credentials never have pem).
+      if (desired.length === 0) continue;
 
       // Warn if duplicate names exist in Auth0 — we can't match safely
       const existingNames = existing.map((c) => c.name);
--- a/src/tools/auth0/handlers/clientAuthCredentials.ts
+++ b/src/tools/auth0/handlers/clientAuthCredentials.ts
@@ -47,17 +47,24 @@ export default class ClientAuthCredentialsHandler {
     const allowDelete =
       this.config('AUTH0_ALLOW_DELETE') === true || this.config('AUTH0_ALLOW_DELETE') === 'true';
 
-    // Fetch existing clients from Auth0 so we can resolve names to IDs
-    // when client_id is null (directory mode strips it from exported JSON)
-    const existingClients = await paginate(this.client.clients.list, {
-      paginate: true,
-      is_global: false,
-    });
-    const clientIdByName = new Map(existingClients.map((c: any) => [c.name, c.client_id]));
+    const needsNameResolution = clients.some(
+      (c) => !c.client_id && c.client_authentication_methods
+    );
+
+    // Only fetch all clients when directory mode has stripped client_id (null) from any entry
+    // that also has client_authentication_methods — avoids an extra API call in the common case.
+    let clientIdByName = new Map<string, string>();
+    if (needsNameResolution) {
+      const existingClients = await paginate(this.client.clients.list, {
+        paginate: true,
+        is_global: false,
+      });
+      clientIdByName = new Map(existingClients.map((c: any) => [c.name, c.client_id]));
+    }
 
     for (const client of clients) {
       if (!client.client_authentication_methods) continue;
-      const clientId = (client.client_id || clientIdByName.get(client.name)) as string | undefined;
+      const clientId = client.client_id || (client.name && clientIdByName.get(client.name));
       if (!clientId) continue;
       client.client_id = clientId;
 
@@ -140,6 +147,7 @@ export default class ClientAuthCredentialsHandler {
       const createdIds = [...createdIdByName.values()];
       const finalIds = [...keptIds, ...createdIds];
 
+      let updatedAuthMethods: Record<string, any> = {};
       if (finalIds.length > 0) {
         // Build method→ids map. For kept credentials use Auth0's stored type to infer method;
         // for created credentials use the method from the user's config entry (by name).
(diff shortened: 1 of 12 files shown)


src/tools/auth0/handlers/clientAuthCredentials.ts
```diff
@@ -0,0 +1,227 @@
...
+        existing = (await this.client.clients.credentials.list(clientId)) as any[];
+      } catch (err) {
+        log.warn(
+          `clientAuthCredentials: failed to list credentials for client "${clientName}": ${err}`
+        );
+        continue;
+      }
+
+      // If no pem-based credentials in config and nothing in Auth0, nothing to do
+      if (desired.length === 0 && existing.length === 0) continue;
```
ankita10119: If desired is empty but existing is not (e.g. after a plain auth0 export where credentials land in config with name but no pem), won't toDelete end up containing all existing credentials, causing them to be deleted on the next deploy with `AUTH0_ALLOW_DELETE=true`? Should this be if `(desired.length === 0) continue`; to skip all reconciliation when no pem is present in config?

ankita10119 (changes requested):

src/tools/auth0/handlers/clientAuthCredentials.ts
harshithRai: agreed! that's a valid catch.
Updated to skip reconciliation when no pem is present in config

ankita10119 (approved):

mendhak:
Thanks for this, it's useful to have it part of the main configuration now! 

### PEM format

Future searchers, unfortunately there seems to be very little in the way of documentation for the 'pem' format that's required here. I tried searching around but couldn't find anything so I hope this helps others. This format is what worked for me:

```
"pem": "-----BEGIN PUBLIC KEY-----\nMIIB...\nXXXXX\n....AQAB\n-----END PUBLIC KEY-----\n",
```

I used these steps to generate a key and get it into a single line

```
openssl genrsa -out test_key.pem 2048
openssl rsa -in test_key.pem -outform PEM -pubout -out test_key.pem.pub

# to use it in the json:
awk '{printf "%s\\n", $0}' test_key.pem.pub  
```

### Replacing requires deleting then create (or run twice)

In my testing I noticed that if you have 2 keys in place already, then replacing one key wouldn't result in an automatic delete then recreate. It first errors as it tries to create a third key and fails (maximum 2 allowed), and after that deletes the key that was removed. Then on rerunning the pipeline the new 2nd key is created. Something to keep in mind when using this, I assume the intended workflow is to first delete the old key then create the new one (or just run it twice)

PR #1421, opened 2026-07-08
feat: add third_party_client_access support for organizations

ankita10119:
### 🔧 Changes

Adds `third_party_client_access` (enum: `allow` | `block`) to the organizations JSON schema in support of the Third Party Apps GA feature.                                
                                                                                                                                                                            
This is a top-level property on the organization entity that controls whether third-party clients can perform standard user flows (Authorization Code, Refresh Token exchange) within the context of the organization. The property is gated by the `organizations_third_party_client_support` feature flag on the tenant.                     
                                                                                                                                                                            
No handler logic changes were required, the existing create, update, and export flows pass top-level org properties through to the Management API automatically. The schema addition ensures the property is accepted during validation and surfaced correctly in dry-run previews.

### 📚 References



### 🔬 Testing

Tested manually against a tenant with `organizations_third_party_client_support` enabled:

  - **Import (`allow`)**: org created with `third_party_client_access: "allow"` confirmed via Management API
  - **Export**: `third_party_client_access` present on all orgs in exported YAML
  - **Dry-run**: changing `allow` → `block` correctly surfaces as `UPDATE` in the preview table
  - **Import (update)**: PATCH applied, value updated to `block` confirmed via Management API
  - **Schema validation**: invalid value (e.g. `invalid_value`) rejected before any API call with `should be equal to one of the allowed values`

Unit test suite (1304 tests) passes with no failures.

### 📝 Checklist

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 22.14.0
+          node-version: 24.x
           registry-url: https://registry.npmjs.org
           package-manager-cache: false
 
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 24.x
+          node-version: 22.14.0
           registry-url: https://registry.npmjs.org
           package-manager-cache: false
 
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "typescript": "~5.9.3"
   },
   "engines": {
-    "node": ">=24.0.0"
+    "node": ">=20.19.0"
   },
   "keywords": [
     "auth0",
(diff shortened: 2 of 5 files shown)


src/tools/auth0/handlers/organizations.ts
```diff
@@ -87,6 +87,10 @@ export const schema = {
           required: ['domain', 'status'],
         },
       },
+      third_party_client_access: {
+        type: 'string',
+        enum: ['allow', 'block'],
```
kushalshit27: nit: Use enum from Management lib

kushalshit27 (dismissed):

package.json
```diff
@@ -80,7 +80,7 @@
     "typescript": "~5.9.3"
   },
   "engines": {
-    "node": ">=20.19.0"
+    "node": ">=24.0.0"
```
kushalshit27: It's the major version update. Do we need it?

https://github.com/auth0/node-auth0/blob/a823058bd14f6667877bffb625aaa2eadd9d1604/package.json#L113

.github/workflows/npm-publish.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 22.14.0
+          node-version: 24.x
```
kushalshit27: No change should be needed if not a major version

kushalshit27 (changes requested):

kushalshit27 (approved):

PR #1424, opened 2026-07-09
fix: narrow PKJ/mTLS export shape and handle credential deletion edge cases

harshithRai:
### 🔧 Changes

Follow-up changes for client credential lifecycle management introduced in #1417.

**Export shape**:

Narrowed credential export to name + credential_type only; `id`, `kid`, `alg` are no longer included. These fields are Auth0-internal and add noise without value in a deploy-ready config.

If no named credentials exist for a client, `client_authentication_methods` is omitted entirely from the export rather than exported as an empty object.

**YAML**: 
Before:
```yaml
credentials:
  - id: cred_abc123
    name: my-key
    credential_type: public_key
    kid: 9P8IC1...
    alg: RS256
```

After:

```yaml
credentials:
  - name: my-key
    credential_type: public_key
```

**JSON**

Before:
```json
"credentials": [
    { 
       "id": "cred_abc123", 
       "name": "my-key", 
       "credential_type": "public_key", 
       "kid": "9P8IC1...",
       "alg": "RS256" 
    }
]
```

After:
```json
"credentials": [
    { 
        "name": "my-key", 
        "credential_type": "public_key" 
    }
]
```

### 📚 References

Follow-up to #1417 

### 🔬 Testing

Manual:
* Export shape: verified only `name` and `credential_type` appear in exported `YAML`, `id`/`kid`/`alg` no longer present
* Empty export: verified `client_authentication_methods` is omitted entirely when no named credentials exist
* Absent = deletion: removed `client_authentication_methods` from config, verified all credentials deleted in Auth0 with `AUTH0_ALLOW_DELETE=true`

Unit tests: 
* Updated `clientAuthCredentials` and `clients` handler tests to cover all new behaviors

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/clientAuthCredentials.ts
+++ b/src/tools/auth0/handlers/clientAuthCredentials.ts
@@ -98,8 +98,10 @@ export default class ClientAuthCredentialsHandler {
         continue;
       }
 
-      // If no pem-based credentials in config and nothing in Auth0, nothing to do
-      if (desired.length === 0 && existing.length === 0) continue;
+      // Skip reconciliation entirely when no pem is present — pem is the opt-in signal.
+      // Without this guard, a plain export→deploy would delete all existing credentials
+      // because desired would be empty (exported credentials never have pem).
+      if (desired.length === 0) continue;
 
       // Warn if duplicate names exist in Auth0 — we can't match safely
       const existingNames = existing.map((c) => c.name);
--- a/src/tools/auth0/handlers/clientAuthCredentials.ts
+++ b/src/tools/auth0/handlers/clientAuthCredentials.ts
@@ -47,17 +47,24 @@ export default class ClientAuthCredentialsHandler {
     const allowDelete =
       this.config('AUTH0_ALLOW_DELETE') === true || this.config('AUTH0_ALLOW_DELETE') === 'true';
 
-    // Fetch existing clients from Auth0 so we can resolve names to IDs
-    // when client_id is null (directory mode strips it from exported JSON)
-    const existingClients = await paginate(this.client.clients.list, {
-      paginate: true,
-      is_global: false,
-    });
-    const clientIdByName = new Map(existingClients.map((c: any) => [c.name, c.client_id]));
+    const needsNameResolution = clients.some(
+      (c) => !c.client_id && c.client_authentication_methods
+    );
+
+    // Only fetch all clients when directory mode has stripped client_id (null) from any entry
+    // that also has client_authentication_methods — avoids an extra API call in the common case.
+    let clientIdByName = new Map<string, string>();
+    if (needsNameResolution) {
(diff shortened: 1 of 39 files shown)


src/tools/auth0/handlers/clients.ts
```diff
@@ -650,17 +650,15 @@ export default class ClientHandler extends DefaultAPIHandler {
...
-          }
-        }
+        // Always strip client_authentication_methods from the PATCH payload —
+        // the clientAuthCredentials handler (order 70) owns this field entirely.
+        // Auth0 rejects credential objects with name/credential_type in PATCH.
+        // Also always strip token_endpoint_auth_method — Auth0 rejects it while
+        // client_authentication_methods is active, and we cannot know at this point
+        // whether the existing Auth0 state has credentials. The clientAuthCredentials
+        // handler (order 70) sets token_endpoint_auth_method when clearing credentials.
+        delete item.token_endpoint_auth_method;
```
ankita10119: Does this strip `token_endpoint_auth_method` from all client PATCH payloads, including clients that have no `PKJ/mTLS` credentials configured,   meaning users who explicitly set this field in their config will have it silently ignored on deploy?

src/tools/auth0/handlers/clientAuthCredentials.ts
```diff
@@ -47,12 +47,9 @@ export default class ClientAuthCredentialsHandler {
     const allowDelete =
       this.config('AUTH0_ALLOW_DELETE') === true || this.config('AUTH0_ALLOW_DELETE') === 'true';
 
-    const needsNameResolution = clients.some(
-      (c) => !c.client_id && c.client_authentication_methods
-    );
+    const needsNameResolution = clients.some((c) => !c.client_id);
```
ankita10119: In directory mode` client_id` is null for every client, so this is always true, won't this cause a paginated `GET /clients` call and a `GET /clients/{id}/credentials` call for every client on every directory-mode deploy, even for tenants with no `PKJ/mTLS` credentials at all?

src/tools/auth0/handlers/clientAuthCredentials.ts
```diff
@@ -98,10 +105,8 @@ export default class ClientAuthCredentialsHandler {
         continue;
       }
 
-      // Skip reconciliation entirely when no pem is present — pem is the opt-in signal.
-      // Without this guard, a plain export→deploy would delete all existing credentials
-      // because desired would be empty (exported credentials never have pem).
-      if (desired.length === 0) continue;
+      // No credentials in Auth0 and none desired — nothing to do.
+      if (existing.length === 0 && desired.length === 0) continue;
```
ankita10119, line 109: nit: We might not need to check `desired.length === 0` again here, as we have already checked above

ankita10119 (approved):

PR #1429, opened 2026-07-22
fix: warn and skip deprecated guardian phone provider on 403 during import

harshithRai:
### 🔧 Changes

Importing a YAML configuration exported from an older tenant into a newer tenant could abort the entire import when the legacy per-factor phone provider API is disabled on the target tenant. The Management API returns `403 legacy_mfa_phone_provider_not_allowed` ("Insufficient privileges to use this deprecated feature"), and the import failed hard during the `processChanges` stage.

The export/`getType` path already handled such 403s gracefully via `isForbiddenFeatureError` (warn + skip). This change wires the same behavior into the import/`processChanges` path for the guardian phone resources, so a deprecated/forbidden feature now logs a warning and is skipped instead of failing the deployment.

- `guardianFactorTemplates`, `guardianPhoneFactorSelectedProvider`, `guardianPhoneFactorMessageTypes`: wrapped the write API calls in `processChanges` to catch a deprecated/forbidden 403 and skip instead of throwing.
- `isForbiddenFeatureError` (utils): cleaned up the warning to read the human-readable `message` and `errorCode` from the response body (`originalError.response.body`) instead of logging the full serialized error blob. This improves logging for every 403-skippable resource that uses the helper.
- `guardianPhoneFactorSelectedProvider`, `guardianPhoneFactorMessageTypes`: simplified `isFeatureUnavailableError` to only handle the 404 (endpoint-not-available) case. The `hooks_not_allowed` / `voice_mfa_not_allowed` 403 branches were redundant since `isForbiddenFeatureError` already catches any 403 - these 403s now emit a warn log instead of being skipped silently, consistent with other forbidden-feature handling.

### 📚 References

- Error `legacy_mfa_phone_provider_not_allowed` is returned at runtime by the Management API but is **not documented** in the OpenAPI spec or typed in auth0-node (v6.x) - verified against `management-api-oas.json`. The fix therefore treats any 403 on these guardian phone endpoints as a skippable deprecated/forbidden feature rather than matching a specific errorCode string.
- `GET` / `PUT` [`/guardian/factors/sms/templates`](https://auth0.com/docs/api/management/v2/guardian/put-factor-sms-templates) are documented as deprecated in favor of the phone-templates endpoints: https://auth0.com/docs/api/management/v2/guardian/put-factor-phone-templates

### 🔬 Testing

Verified end-to-end against a live tenant with **"Use Tenant-Level Messaging Provider"** enabled (which disables the legacy per-factor phone API):

- **Before the fix:** `a0deploy import` with `guardianPhoneFactorSelectedProvider` in the YAML failed with `Problem running command import during stage processChanges when processing type guardianPhoneFactorSelectedProvider` and exited with code `1`.
- **After the fix:** the same import logged `warn: Insufficient privileges to use this deprecated feature. (legacy_mfa_phone_provider_not_allowed) - Skipping guardianPhoneFactorSelectedProvider`, reported `Import Successful`, and exited with code `0`.

Unit tests added/updated:
- `guardianFactorTemplates`: verifies `processChanges` warns and skips on a legacy-feature 403 instead of throwing.
- `isForbiddenFeatureError`: verifies the cleaned-up warning message and errorCode are extracted from the response body.
- `guardianPhoneFactorSelectedProvider` / `guardianPhoneFactorMessageTypes`: existing "endpoint is disabled for tenant" tests still pass, now routing the 403 through `isForbiddenFeatureError` (returns null/skips, with a warn log).

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/guardianPhoneFactorSelectedProvider.ts
+++ b/src/tools/auth0/handlers/guardianPhoneFactorSelectedProvider.ts
@@ -20,16 +20,7 @@ const isFeatureUnavailableError = (err) => {
     // Older Management API version where the endpoint is not available.
     return true;
   }
-  if (
-    err.statusCode === 403 &&
-    err.originalError &&
-    err.originalError.response &&
-    err.originalError.response.body &&
-    err.originalError.response.body.errorCode === 'hooks_not_allowed'
-  ) {
-    // Recent Management API version, but with feature explicitly disabled.
-    return true;
-  }
+  // 403s (feature explicitly disabled) are handled by isForbiddenFeatureError.
   return false;
 };
 
--- a/src/tools/auth0/handlers/guardianPhoneFactorSelectedProvider.ts
+++ b/src/tools/auth0/handlers/guardianPhoneFactorSelectedProvider.ts
@@ -81,9 +81,17 @@ export default class GuardianPhoneSelectedProviderHandler extends DefaultHandler
     }
 
     const data = guardianPhoneFactorSelectedProvider;
-    await this.client.guardian.factors.phone.setProvider(
-      data as Management.SetGuardianFactorsProviderPhoneRequestContent
-    );
+    try {
+      await this.client.guardian.factors.phone.setProvider(
+        data as Management.SetGuardianFactorsProviderPhoneRequestContent
+      );
+    } catch (err) {
+      if (isFeatureUnavailableError(err) || isForbiddenFeatureError(err, this.type)) {
+        // Feature is deprecated/disabled on this tenant; warn and skip instead of failing the import.
+        return;
+      }
+      throw err;
+    }
     this.updated += 1;
     this.didUpdate(guardianPhoneFactorSelectedProvider);
   }
--- a/src/tools/auth0/handlers/guardianFactorTemplates.ts
+++ b/src/tools/auth0/handlers/guardianFactorTemplates.ts
@@ -67,14 +67,22 @@ export default class GuardianFactorTemplatesHandler extends DefaultHandler {
       guardianFactorTemplates.map(async (fatorTemplates) => {
         const { name, ...data } = fatorTemplates;
         const params = { name: fatorTemplates.name };
-        if (name === 'sms') {
-          await this.client.guardian.factors.sms.setTemplates(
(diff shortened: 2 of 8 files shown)


src/tools/auth0/handlers/guardianPhoneFactorSelectedProvider.ts
```diff
@@ -81,9 +81,17 @@ export default class GuardianPhoneSelectedProviderHandler extends DefaultHandler
...
     const data = guardianPhoneFactorSelectedProvider;
-    await this.client.guardian.factors.phone.setProvider(
-      data as Management.SetGuardianFactorsProviderPhoneRequestContent
-    );
+    try {
+      await this.client.guardian.factors.phone.setProvider(
+        data as Management.SetGuardianFactorsProviderPhoneRequestContent
+      );
+    } catch (err) {
+      if (isFeatureUnavailableError(err) || isForbiddenFeatureError(err, this.type)) {
```
ankita10119, line 80: Hey, curious about the `hooks_not_allowed `errorCode here, is this specific to the phone provider endpoint, or could it have been carried over from somewhere else? 

Asking because I'd expect Auth0 to return something like `legacy_mfa_phone_provider_not_allowed` here rather than a hooks related errorCode                                     
                                                                                                                                                                            
If it's not needed, we could simplify `isFeatureUnavailableError` to just handle the 404 case and let `isForbiddenFeatureError` take care of 403s (which it already does)   

Totally a nit though, not a blocker!

harshithRai: Good catch! you're right. `hooks_not_allowed` was pre-existing and redundant since `isForbiddenFeatureError` already handles any 403. Simplified `isFeatureUnavailableError` to just the 404 case in both handlers. Pushed the change 👍

ankita10119:
The new `processChanges` catch blocks in `guardianPhoneFactorMessageTypes` and `guardianPhoneFactorSelectedProvider` don't have test coverage, **Codecov** is flagging 3 missing
lines in each 

Could you add 403 tests for both, similar to the one already added for `guardianFactorTemplates`?

ankita10119 (approved):

PR #1430, opened 2026-07-23
feat: add token_vault_privileged_access support for clients

harshithRai:
### 🔧 Changes

Adds support for the `token_vault_privileged_access` property on clients, which hardens a privileged Token Vault worker by restricting the caller IPs, connections, and scopes it may use at runtime.

- **`clients` schema** - new `token_vault_privileged_access` object:
  - `ip_allowlist` (array of strings) - IPv4/IPv6 addresses or CIDR ranges permitted to call token exchange on behalf of the client.
  - `grants` (array of objects) - connection/scope pin objects, each with `connection` (name) and `scopes` (array). Max 5 connections; max 20 scopes total (enforced server-side).
- **Export-only field.** `token_vault_privileged_access` is exported for visibility but is **not deployed** - it is stripped from create/update payloads. The Management API requires a `credentials` array (tenant-specific credential `id` references) whenever the object is sent; those ids are never persisted by the CLI because they are not portable across tenants. Sending the object without them fails validation (`Missing required property: credentials`), and sending exported ids would re-send stale references on a cross-tenant deploy. Manage `token_vault_privileged_access` directly on the tenant.
- No `node-auth0` bump required - the installed SDK (v6.0.0) already ships these types.

**Early Access:** `token_vault_privileged_access` requires the `token_vault_subject_type_jwt_ea_rollout` feature flag on the tenant, and writes additionally require the `create:/update:client_token_vault_privileged_access` scopes. Since the field is export-only in the CLI, these requirements affect only manual configuration on the tenant.

#### Exported shape

`credentials` is stripped; `ip_allowlist` and `grants` are kept for visibility.

YAML:

```yaml
clients:
  - name: My Token Vault Privileged App
    app_type: non_interactive
    token_vault_privileged_access:
      ip_allowlist:
        - "192.168.1.0/24"
        - "10.0.0.1"
      grants:
        - connection: google-oauth2
          scopes:
            - "https://www.googleapis.com/auth/calendar.readonly"
        - connection: slack
          scopes:
            - "chat:write"
            - "channels:read"
```

JSON (directory format):

```json
{
  "name": "My Token Vault Privileged App",
  "app_type": "non_interactive",
  "token_vault_privileged_access": {
    "ip_allowlist": ["192.168.1.0/24", "10.0.0.1"],
    "grants": [
      {
        "connection": "google-oauth2",
        "scopes": ["https://www.googleapis.com/auth/calendar.readonly"]
      },
      { "connection": "slack", "scopes": ["chat:write", "channels:read"] }
    ]
  }
}
```

### 🔬 Testing

**Unit Tests:**
in `test/tools/auth0/handlers/clients.tests.js`:
- Schema validation accepts a valid `token_vault_privileged_access` shape.
- Export strips `credentials` while keeping `ip_allowlist`/`grants`.
- Create/update payloads strip the entire `token_vault_privileged_access` object.

**Manual Tests:**
Verified end-to-end on a live EA-enabled tenant: configured a privileged worker, ran export→import, and confirmed the exported file omits `credentials`, the import completes with no validation error, and the worker's on-tenant config is left intact.

### 📝 Checklist

### DIFF
--- a/docs/resource-specific-documentation.md
+++ b/docs/resource-specific-documentation.md
@@ -1705,10 +1705,10 @@ Contents of `My API Client.json` (deploy-time, with pem):
 
 The Deploy CLI supports the `token_vault_privileged_access` property on clients, which hardens a privileged Token Vault worker by restricting the caller IPs, connections, and scopes it may use at runtime.
 
-| Field          | Type              | Description                                                                                                                           |
-| -------------- | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
-| `ip_allowlist` | array of strings  | IPv4/IPv6 addresses or CIDR ranges permitted to call token exchange on behalf of this client.                                         |
-| `grants`       | array of objects  | Connection/scope pin objects. Each has a `connection` (name) and `scopes` (array). Max 5 connections; max 20 scopes total.             |
+| Field          | Type             | Description                                                                                                                |
+| -------------- | ---------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `ip_allowlist` | array of strings | IPv4/IPv6 addresses or CIDR ranges permitted to call token exchange on behalf of this client.                              |
+| `grants`       | array of objects | Connection/scope pin objects. Each has a `connection` (name) and `scopes` (array). Max 5 connections; max 20 scopes total. |
 
 ```yaml
 clients:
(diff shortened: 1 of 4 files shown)


docs/resource-specific-documentation.md
```diff
@@ -1698,3 +1698,34 @@ Contents of `My API Client.json` (deploy-time, with pem):
...
+        - connection: google-oauth2
+          scopes:
+            - 'https://www.googleapis.com/auth/calendar.readonly'
+        - connection: slack
+          scopes:
+            - 'chat:write'
+            - 'channels:read'
+```
+
+> **Note:** The `credentials` sub-object of `token_vault_privileged_access` is **not** managed by the Deploy CLI. On read, Auth0 returns it as credential references (`id`) rather than names or key material, and the Deploy CLI never syncs IDs. Manage the privileged client's credentials directly on the tenant.
```
kushalshit27: `token_vault_privileged_access.credentials` is never stripped on export, so an export→deploy round-trip re-sends stale credential IDs.

harshithRai: Fixed. You're right that the original approach re-sent stale credential IDs on round-trip.
Made the field export-only: credentials stripped on export, and the whole object stripped on create/update - so nothing stale is ever re-sent. Verified end-to-end on an EA tenant; tests + docs updated.

kushalshit27 (approved):

PR #1434, opened 2026-07-24
fix: handle 403 entitlement errors for bot detection and captcha on export                                                                                                

ankita10119:
### 🔧 Changes

Fixes 403 entitlement error handling for bot detection and captcha on export, in preparation for ROAD-8051 (`free_bot_detection_deprecation` FF).                           
                                                                                                                                                                            
  Export (GET):                                                                                                                                                           
  - Decoupled `botDetection.get()` and `captcha.get()` into independent calls with independent error handling. Previously both were in a single `Promise.all`, a 403 on bot detection would silently discard captcha data even though captcha returned 200.                                                                                           
  - A 403 on either endpoint now logs a warning and skips that resource independently, without affecting the other.                                                         
  - Updated warning message to reference the tenant's plan rather than OAuth scopes.

  Deploy (PATCH):
  - Bot detection and captcha 403s surface to the user as errors (per API team confirmation that PATCH 403 should not be handled gracefully).

### 📚 References


### 🔬 Testing

Unit tests cover:
  - GET 403 on bot detection -  captcha data is preserved in the export
  - GET 403 on captcha - bot detection data is preserved in the export
  - Happy path export and import verified on real tenant

E2E verified against a real tenant with `free_bot_detection_deprecation` FF enabled and bot detection entitlement revoked:
  - Both `botDetection` and `captcha` 403s handled gracefully with correct warning messages
  - Export succeeds without those fields in the output
  - `breachedPasswordDetection`, `bruteForceProtection`, `suspiciousIpThrottling` unaffected

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/attackProtection.ts
+++ b/src/tools/auth0/handlers/attackProtection.ts
@@ -258,15 +258,26 @@ export default class AttackProtectionHandler extends DefaultAPIHandler {
     let captcha: Asset | null = null;
 
     try {
-      [botDetection, captcha] = await Promise.all([
-        this.client.attackProtection.botDetection.get(),
-        this.client.attackProtection.captcha.get(),
-      ]);
+      botDetection = await this.client.attackProtection.botDetection.get();
     } catch (err) {
       if (err.statusCode === 403) {
         log.warn(
-          'Bot Detection API are not enabled for this tenant. Please verify `scope` or contact Auth0 support to enable this feature.'
+          'Bot detection is not available on this tenant\'s current plan. Skipping bot detection settings.'
         );
+      } else {
+        throw err;
+      }
+    }
+
+    try {
+      captcha = await this.client.attackProtection.captcha.get();
+    } catch (err) {
+      if (err.statusCode === 403) {
+        log.warn(
+          'Captcha is not available on this tenant\'s current plan. Skipping captcha settings.'
+        );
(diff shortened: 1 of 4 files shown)


src/tools/auth0/handlers/attackProtection.ts
```diff
@@ -258,15 +258,26 @@ export default class AttackProtectionHandler extends DefaultAPIHandler {
...
-      ]);
+      botDetection = await this.client.attackProtection.botDetection.get();
+    } catch (err) {
+      if (err.statusCode === 403) {
+        log.warn(
+          "Bot detection is not available on this tenant's current plan. Skipping bot detection settings."
+        );
+      } else {
+        throw err;
+      }
```
harshithRai, line 269: Just want to double-check the error handling here. Previously the `catch` swallowed everything (only 403s got logged, other errors were silently ignored), whereas now anything that isn't a 403 re-throws and fails the whole export. 
I think that's indeed the "more correct" behavior, just want to confirm it's intentional rather than a side effect of the refactor. 

If so, might be worth a quick test on the re-throw path since it's the riskiest bit of the change., LGTM otherwise

ankita10119: Addressed and updated

harshithRai (approved):

PR #1436, opened 2026-07-27
fix: strip unresolved placeholders on deploy instead of throwing

ankita10119:
### 🔧 Changes

Previously, deploying a resource with an unresolved `##...##` or `@@...@@` placeholder would throw an error and abort the deploy. This caused the export→import round-trip to fail for connections and other resources where the CLI auto-generates secret placeholders on export (e.g. `##CONNECTIONS_WAAD_SECRET##`), since Auth0 does not expose the actual secret value, making it impossible for users to provide the keyword mapping.
                                                                                                                                                                            
**Change**: Instead of throwing, unresolved placeholders are now stripped from the payload before the API call, with a warn log per field. Auth0 preserves its existing value for any stripped field.                                                                                                                                                   

  - Added `stripUnresolvedPlaceholders` utility in `src/tools/utils.ts`
  - `APIHandler.processChanges` (create, update, conflict paths) now calls `stripUnresolvedPlaceholders` in place of `validateNoUnresolvedPlaceholders`

### 📚 References

- PR #1360 

### 🔬 Testing

 - Unit tests updated in `test/tools/utils.test.js` and `test/tools/auth0/handlers/default.tests.ts`
 - Tested against real tenant: exported a WAAD connection (producing `##CONNECTIONS_WAAD_SECRET##`), ran import without defining the mapping, field was stripped with a warning and the connection updated successfully

### 📝 Checklist

### DIFF
--- a/src/tools/utils.ts
+++ b/src/tools/utils.ts
@@ -316,6 +316,33 @@ export const stripObfuscatedFieldsFromPayload = (
   return newAsset;
 };
 
+// Strips unresolved ##...## / @@...@@ placeholders before sending to Auth0.
+// Keyword replacement runs first; any remaining placeholder means no mapping was
+// provided. The field is dropped so Auth0 preserves its existing value.
+export const stripUnresolvedPlaceholders = (
+  data: Asset | null,
+  resourceType: string,
+  resourceName: string
+): Asset | null => {
+  if (data === null) return data;
+
+  const unresolved = collectUnresolvedPlaceholders(data);
+  if (unresolved.length === 0) return data;
+
+  const newAsset = { ...data };
+  unresolved.forEach(({ path, value }) => {
+    log.warn(
+      `Stripping unresolved placeholder for ${resourceType} "${resourceName}" field "${path}": ${value}. ` +
+        `To use this value, define ${path
+          .split('.')
+          .pop()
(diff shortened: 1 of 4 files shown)


src/tools/utils.ts
```diff
@@ -316,6 +316,33 @@ export const stripObfuscatedFieldsFromPayload = (
...
+  if (unresolved.length === 0) return data;
+
+  const newAsset = { ...data };
+  unresolved.forEach(({ path, value }) => {
+    log.warn(
+      `Stripping unresolved placeholder for ${resourceType} "${resourceName}" field "${path}": ${value}. ` +
+        `To use this value, define ${path
+          .split('.')
+          .pop()
+          ?.toUpperCase()} in AUTH0_KEYWORD_REPLACE_MAPPINGS. The existing value on Auth0 will be preserved.`
```
harshithRai: I think the suggested key here might be slightly off. It's derived from the field name rather than the placeholder itself. For `options.client_secret: ##CONNECTIONS_WAAD_SECRET##`, it would tell the user to define `CLIENT_SECRET`, but the value they actually need to map is `CONNECTIONS_WAAD_SECRET`, so following the message as-is wouldn't quite work. Could we maybe pull the name from `value` instead (stripping the `##`/`@@`)? 

It's already available right there in the loop, so it might be a small change.

ankita10119: Addressed and updated

harshithRai (approved):

PR #1437, opened 2026-07-27
fix: resolve spurious dry-run diffs for rules, themes, and hooks

ankita10119:
### 🔧 Changes

Fixes three sources of spurious dry-run diffs that caused a clean export-then-dry-run cycle to propose changes when none existed.                                                                                                           
                                         
  **Rules** (`handlers/rules.ts`)                                                                                                                                           
`dryRunChanges` was delegating to `calcChanges`, which internally runs the `duplicateItems` suffix-generation path. This caused every rule in the tenant to appear twice in the diff, once under its real name and once with a random suffix (e.g. `Add voucherID to AccessToken-betv2`).
Fixed by calling `calculateDryRunChanges` directly, which performs a clean name-based comparison without any renaming side-effects.

  **Themes** (`handlers/themes.ts`)
Themes are singletons, but a default export strips `themeId` from the local file (because `AUTH0_EXPORT_IDENTIFIERS` defaults to `false`).
Without a `themeId`, the dry-run matcher found no match for the local theme and proposed a CREATE + DELETE on every run. Fixed by adding a `dryRunChanges` override that backfills `themeId` from the remote singleton before comparison, so the matcher resolves correctly without requiring `AUTH0_EXPORT_IDENTIFIERS`.

  **Hooks** (`handlers/hooks.ts`)
`getType` was destructuring `{ data: secrets }` from the secrets response, but the SDK returns the object directly. This meant secrets were always `undefined` locally, causing hooks with secrets to appear changed on every dry run. Fixed by removing the `.data` unwrap.

### 🔬 Testing

Unit tests added/extended for all three fixes:

  - `test/tools/auth0/handlers/dryRun.tests.ts` - new rules clean-cycle test (no creates/updates/deletes after a clean export), and two new themes regression tests (no themeId → no create/delete; content diff → update only)
  - `test/tools/auth0/handlers/themes.tests.js` - new`#themes dryRunChanges` describe block with 5 cases covering the
    backfill logic, content-diff detection, and null-guard behaviour

All 37 affected tests pass (`npm test`).

For end-to-end validation: export any tenant, immediately run`--dry-run` against the same tenant with the same exported file - rules, themes, and hooks should all show zero proposed changes.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/themes.ts
+++ b/src/tools/auth0/handlers/themes.ts
@@ -483,11 +483,17 @@ export default class ThemesHandler extends DefaultHandler {
 
     const existing = await this.getType();
 
+    // getType() returns null when Universal Login customizations are not enabled
+    // on the tenant (the 400 "no-code not enabled" path in getThemes).
+    if (existing === null) {
+      return { del: [], create: [], conflicts: [], update: [] };
+    }
+
     // Themes are singletons — at most one per tenant. If the local theme has no
     // themeId (stripped during export when --export_ids is not set), backfill it
     // from the remote so the matcher can find a match without requiring --export_ids.
     const normalizedThemes = themes.map((theme) => {
-      if (!theme.themeId && existing && existing.length > 0 && existing[0].themeId) {
+      if (!theme.themeId && existing.length > 0 && existing[0].themeId) {
         return { ...theme, themeId: existing[0].themeId };
       }
       return theme;
@@ -496,7 +502,6 @@ export default class ThemesHandler extends DefaultHandler {
     return calculateDryRunChanges({
       type: this.type,
       assets: normalizedThemes,
-      // @ts-ignore
       existing,
       identifiers: this.identifiers,
       ignoreDryRunFields: this.getEffectiveIgnoreDryRunFields(),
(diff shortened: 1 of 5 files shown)


src/tools/auth0/handlers/themes.ts
```diff
@@ -437,6 +438,35 @@ export default class ThemesHandler extends DefaultHandler {
...
   }
 
+  async dryRunChanges(assets: Assets): Promise<CalculatedChanges> {
+    const { themes } = assets;
+
+    if (!themes) {
+      return { del: [], create: [], conflicts: [], update: [] };
+    }
+
+    const existing = await this.getType();
```
harshithRai, line 484: Small thing: `getType()` returns `null` when no-code isn't enabled (the 400 case in `getThemes`), and passing that straight into `calculateDryRunChanges` will blow up on `[null]`. The rules fix in this same PR guards it - could we do the same here?

```ts
const existing = await this.getType();
if (existing === null) {
  return { del: [], create: [], conflicts: [], update: [] };
}
```

ankita10119: Addressed and updated

harshithRai (approved):

PR #1438, opened 2026-07-28
fix: prevent dry-run crash on clientAuthCredentials handler

harshithRai:
### 🔧 Changes

Fixes a crash during `import --dry_run` that aborted the entire dry-run preview with `handler.dryRunChanges is not a function`.

The `clientAuthCredentials` feature added two handlers that were implemented for the deploy path only and did not define a `dryRunChanges()` method. Because the dry-run engine calls `dryRunChanges()` on every registered handler, the run crashed as soon as it reached this resource - regardless of whether the tenant actually used client credentials.

- Added a `dryRunChanges()` method to `ClientAuthCredentialsHandler` and `ClientAuthCredentialsPreHandler`, each returning empty changes (`{ del: [], create: [], conflicts: [], update: [] }`) - the same no-op shape the base handler returns when there is nothing to do.
- Client credentials are per-client sub-resources with no top-level asset list, so they are intentionally not itemized in the dry-run diff for now. Credential changes are still applied correctly during a real `import`.
- Hardened the dry-run loop to skip any handler that doesn't implement `dryRunChanges` (with a warning) instead of crashing, so a future deploy-only handler can't reintroduce this failure.
- A normal `import` (without `--dry_run`) was never affected.

**No changes to the YAML/JSON config shape** - the `clientAuthCredentials` / `client_authentication_methods` structure is unchanged. This only affects dry-run behavior.

### 🔭 Future scope

- Itemize client credential creates/deletes in the dry-run diff so the preview reflects the changes a real import would apply, rather than showing no-op.

#### Unit Test
- Added regression tests confirming both handlers return empty changes without crashing during dry run.
- Added a conformance test asserting every registered handler implements `dryRunChanges`, so a non-conforming handler is caught at build time.

#### Manual:
- `a0deploy import --input_file=./tenant.yaml --dry_run` now completes on a tenant with client credentials instead of crashing.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/clientAuthCredentials.ts
+++ b/src/tools/auth0/handlers/clientAuthCredentials.ts
@@ -1,5 +1,5 @@
 import { order } from './default';
-import { Assets, Auth0APIClient } from '../../../types';
+import { Assets, Auth0APIClient, CalculatedChanges } from '../../../types';
 import { ConfigFunction } from '../../../configFactory';
 import { paginate } from '../client';
 import log from '../../../logger';
@@ -37,6 +37,14 @@ export default class ClientAuthCredentialsHandler {
 
   async validate() {
     return null;
+  }
+
+  // Credentials are per-client sub-resources with no top-level asset list, so they are not
+  // modeled in the dry-run diff. Return empty changes so the dry-run loop (which calls
+  // dryRunChanges on every handler) does not crash. Credential changes are still applied
+  // during a real import via processChanges.
+  async dryRunChanges(_assets: Assets): Promise<CalculatedChanges> {
+    return { del: [], create: [], conflicts: [], update: [] };
   }
 
   @order('70')
--- a/src/tools/auth0/handlers/clientAuthCredentialsPre.ts
+++ b/src/tools/auth0/handlers/clientAuthCredentialsPre.ts
@@ -1,5 +1,5 @@
 import { order } from './default';
-import { Assets, Auth0APIClient } from '../../../types';
+import { Assets, Auth0APIClient, CalculatedChanges } from '../../../types';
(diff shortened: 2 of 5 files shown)


src/tools/auth0/handlers/clientAuthCredentials.ts
```diff
@@ -39,6 +39,14 @@ export default class ClientAuthCredentialsHandler {
     return null;
   }
 
+  // Credentials are per-client sub-resources with no top-level asset list, so they are not
+  // modeled in the dry-run diff. Return empty changes so the dry-run loop (which calls
+  // dryRunChanges on every handler) does not crash. Credential changes are still applied
+  // during a real import via processChanges.
+  async dryRunChanges(_assets: Assets): Promise<CalculatedChanges> {
+    return { del: [], create: [], conflicts: [], update: [] };
```
ankita10119, line 47: This silently hides credential changes from the dry-run preview, if a user's config adds or removes `client_authentication_methods`, dry-run will show no diff but the real import will still apply those changes

That's misleading for anyone using dry-run to audit before deploying                                                                           
                                                                                                                                                                                     
The crash fix is correct and good to merge. Worth tracking a follow-up to surface actual credential diffs in the preview

ankita10119 (approved):

src/tools/auth0/handlers/clientAuthCredentials.ts
harshithRai: Agreed! This is a known, intentional trade-off of the crash fix (called out in the PR description and the code comments): credential changes aren't itemized in the dry-run preview yet, though a real import still applies them. I'll open an internal follow-up for this!

PR #1442, opened 2026-07-29
fix: apply AUTH0_INCLUDED_CONNECTIONS on export

### LINKED ISSUE #1441: AUTH0_INCLUDED_CONNECTIONS has no effect on export, despite docs stating it applies to both export and import
### Checklist

- [x] I have looked into the README and have not found a suitable solution or answer.
- [x] I have looked into the documentation and have not found a suitable solution or answer.
- [x] I have searched the issues and have not found a suitable solution or answer.
- [x] I have upgraded to the latest version of this tool and the issue still persists.
- [x] I have searched the Auth0 Community forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the Auth0 Code of Conduct.

### Description

`AUTH0_INCLUDED_CONNECTIONS` is only applied on `import`. It has no effect on `export`, despite the documentation stating that it affects both.

[`docs/configuring-the-deploy-cli.md`](https://github.com/auth0/auth0-deploy-cli/blob/master/docs/configuring-the-deploy-cli.md#auth0_included_connections) says:

> When configured, only the connections listed by name will be included in **export and import** operations. All other connections in the tenant will be completely ignored.
>
> **Important:** This setting affects all operations (export and import). Connections not in this list will not appear in exports and will not be modified during imports.

The import half is implemented and works. The export half is not implemented in any published version.

The value is read into `assets.include.connections` on both code paths, including inside `dump()`:

```ts
// src/context/directory/index.ts:124-126 (inside dump)
this.assets.include = {
  connections: this.config.AUTH0_INCLUDED_CONNECTIONS || [],
};
```

…but `filterIncluded` has exactly one call site, in the import path:

```ts
// src/tools/auth0/handlers/connections.ts:930-936
async processChanges(assets: Assets): Promise<void> {
  ...
  const includedConnections = (assets.include && assets.include.connections) || [];
  const excludedConnections = (assets.exclude && assets.exclude.connections) || [];
  let changes = await this.calcChanges(assets);
  changes = filterExcluded(changes, excludedConnections);
  changes = filterIncluded(changes, includedConnections);   // only call site
```

`filterIncluded` operates on `CalculatedChanges` (`{del, update, create, conflicts}`), a structure the export path never builds.

Both `dump()` implementations read `assets.exclude` and never `assets.include`:

```ts
// src/context/directory/handlers/connections.ts:69-75
// Filter excluded connections
const excludedConnections = (context.assets.exclude && context.assets.exclude.connections) || [];
if (excludedConnections.length) {
  connections = connections.filter(
    (connection) => !excludedConnections.includes(connection.name)
  );
}
// -> every remaining connection is then written to disk
```

Same in `src/context/yaml/handlers/connections.ts:56-62`.

An exhaustive grep of `src/` on `master` (a8a602f) shows four references total: two writes into `assets.include`, the mutual-exclusion validation against `AUTH0_EXCLUDED_CONNECTIONS`, and one read inside `processChanges`. Identical in `8.29.2`, `8.41.0`, and `9.0.0-beta.1`.

This is a notable asymmetry, since `AUTH0_EXCLUDED_CONNECTIONS` — documented as the deprecated counterpart — *did* gain export filtering, along with folder pruning that preserves excluded connections' files across re-dumps.

### Expectation

With `AUTH0_INCLUDED_CONNECTIONS` configured, `a0deploy export` writes only the listed connections, and leaves connections outside the list unmanaged — mirroring how `AUTH0_EXCLUDED_CONNECTIONS` behaves on export today.

This matters most for the use case the docs call out explicitly — "Managing only specific connections while preserving others (e.g., self-service SSO connections)". A keep-list is the only maintainable shape there: Self-Service SSO mints new connections with generated UUID names over time, so expressing the same intent via `AUTH0_EXCLUDED_CONNECTIONS` requires amending the exclude list every time a connection is created. And the two cannot be combined — `src/context/index.ts:148` throws if both are set.

Concretely, on a tenant with 4 social/enterprise connections we manage plus 27 UUID-named `samlp` connections created by Self-Service SSO, `export` writes all 31 regardless of the include list.

### Reproduction

1. On a tenant with more than one connection, configure only a subset:

   ```json
   {
     "AUTH0_DOMAIN": "<tenant>.us.auth0.com",
     "AUTH0_CLIENT_ID": "<client-id>",
     "AUTH0_INCLUDED_CONNECTIONS": ["google-oauth2", "windowslive"]
   }
   ```

2. Export:

   ```bash
   a0deploy export --config_file=config.json --format=directory --output_folder=local
   ```

3. `local/connections/` contains a `.json` file for **every** connection on the tenant, not just `google-oauth2.json` and `windowslive.json`.

Same result with `--format=yaml` — every connection appears under `connections:` in the output.

`import` correctly restricts changes to the two listed connections, so the discrepancy is export-only.

### Deploy CLI version

8.41.0 (also verified on 8.29.2 and 9.0.0-beta.1)

### Node version

22

### Additional context

Happy to open a PR implementing the export-side filter in both the `directory` and `yaml` handlers, mirroring the existing exclude behaviour — including preserving files of unmanaged connections across re-dumps so the new pruning logic doesn't delete them. Let me know if you'd prefer the fix, or a documentation correction instead.

jithinzac:
### 🔧 Changes

`AUTH0_INCLUDED_CONNECTIONS` was only honoured on `import`. The value was read into `assets.include.connections` on both code paths — including inside `dump()` at `src/context/directory/index.ts:124` — but the only consumer was `filterIncluded` in the connections handler's `processChanges`. `filterIncluded` operates on `CalculatedChanges`, a structure the export path never builds, so exports wrote every connection on the tenant regardless of the include list.

This contradicts the documented behaviour:

> **Important:** This setting affects all operations (export and import). Connections not in this list will not appear in exports and will not be modified during imports.

**`src/context/directory/handlers/connections.ts`** and **`src/context/yaml/handlers/connections.ts`** — `dump()` now filters to the included connections, immediately after the existing exclude filter and mirroring its shape:

```ts
// Filter to included connections
const includedConnections = (context.assets.include && context.assets.include.connections) || [];
if (includedConnections.length) {
  connections = connections.filter((connection) => includedConnections.includes(connection.name));
}
```

**`src/context/directory/handlers/connections.ts`** — pruning is scoped to the listed names when an include list is configured:

```ts
// With an include list configured, connections outside it are unmanaged, so limit pruning
// to the listed names rather than every file in the folder.
const prunableFiles = includedConnections.length
  ? new Set(
      includedConnections.flatMap((name) => [`${sanitize(name)}.json`, `${sanitize(name)}.html`])
    )
  : null;
```

Without this, the folder-pruning logic added alongside export-side exclude filtering would delete the files of every connection outside the include list. Those connections are unmanaged, so their files are preserved across re-dumps — the same guarantee excluded connections already get via the `expectedFiles` seeding. Stale files for connections that *are* in the include list but no longer exist on the tenant are still removed.

No behaviour change when `AUTH0_INCLUDED_CONNECTIONS` is unset: both filters are no-ops on an empty list, and `prunableFiles` stays `null`, leaving the existing prune semantics intact.

### 📚 References

Fixes #1441

The asymmetry this closes: `AUTH0_EXCLUDED_CONNECTIONS`, documented as the deprecated counterpart, gained export filtering while `AUTH0_INCLUDED_CONNECTIONS` did not — and the two cannot be combined, since `src/context/index.ts:148` throws when both are set. That left no way to express the keep-list use case the docs call out explicitly:

> Managing only specific connections while preserving others (e.g., self-service SSO connections, third-party integrations)

A keep-list is the only maintainable shape for that case, because Self-Service SSO mints new connections with generated UUID names over time — expressing the same intent through an exclude list means amending it every time a connection is created.

### 🔬 Testing

Four new tests, placed alongside the existing exclude-dump tests:

`test/context/directory/connections.test.js`

- `should only dump included connections` — a `waad` connection in the list is written; a `samlp` connection outside it is not.
- `should preserve pre-existing files for connections outside the include list on re-dump` — a file written by a prior export survives when its connection is outside the list. Fails without the `prunableFiles` scoping.
- `should remove stale files for included connections no longer present on the tenant` — pruning still works within the include list.

`test/context/yaml/connections.test.js`

- `should only dump included connections` — the returned `connections` array contains just the listed connection.

Full suite: **1342 passing, 1 pending**, no regressions. `npm run build`, `npm run lint` (eslint + kacl), and `npx prettier --check` all clean.

Manual end-to-end verification against a real tenant holding 34 connections — 4 named in `AUTH0_INCLUDED_CONNECTIONS`, the rest mostly UUID-named `samlp` connections created by Self-Service SSO. Identical config file and output folder for both runs, with `AUTH0_INCLUDED_ONLY: ["connections"]`:

- built from `master`: 34 files written to `connections/`
- built from this branch: 4 files — `google-oauth2.json`, `okta-sandbox-oidc-test.json`, `waad.json`, `windowslive.json`

### 📝 Checklist

- [x] I have added documentation for all new/changed functionality (or N/A) — `docs/configuring-the-deploy-cli.md` already documents this behaviour; this change makes the implementation match it, so no doc edit is needed.

### DIFF
--- a/src/context/directory/handlers/connections.ts
+++ b/src/context/directory/handlers/connections.ts
@@ -75,6 +75,12 @@ async function dump(context: DirectoryContext): Promise<void> {
     );
   }
 
+  // Filter to included connections
+  const includedConnections = (context.assets.include && context.assets.include.connections) || [];
+  if (includedConnections.length) {
+    connections = connections.filter((connection) => includedConnections.includes(connection.name));
+  }
+
   const connectionsFolder = path.join(context.filePath, constants.CONNECTIONS_DIRECTORY);
   fs.ensureDirSync(connectionsFolder);
 
@@ -131,11 +137,23 @@ async function dump(context: DirectoryContext): Promise<void> {
     if (dumpedConnection.strategy === 'email') expectedFiles.add(`${connectionName}.html`);
   });
 
+  // With an include list configured, connections outside it are unmanaged, so limit pruning
+  // to the listed names rather than every file in the folder.
+  const prunableFiles = includedConnections.length
+    ? new Set(
+        includedConnections.flatMap((name) => [`${sanitize(name)}.json`, `${sanitize(name)}.html`])
+      )
+    : null;
+
   // Remove files that belong to connections no longer present (and not excluded)
   if (fs.existsSync(connectionsFolder)) {
     for (const existing of fs.readdirSync(connectionsFolder)) {
       const fullPath = path.join(connectionsFolder, existing);
-      if (fs.statSync(fullPath).isFile() && !expectedFiles.has(existing)) {
+      if (
+        fs.statSync(fullPath).isFile() &&
+        !expectedFiles.has(existing) &&
+        (prunableFiles === null || prunableFiles.has(existing))
+      ) {
         fs.removeSync(fullPath);
       }
     }
--- a/src/context/yaml/handlers/connections.ts
+++ b/src/context/yaml/handlers/connections.ts
@@ -61,6 +61,12 @@ async function dump(context: YAMLContext): Promise<ParsedConnections> {
     );
   }
 
+  // Filter to included connections
+  const includedConnections = (context.assets.include && context.assets.include.connections) || [];
+  if (includedConnections.length) {
+    connections = connections.filter((connection) => includedConnections.includes(connection.name));
+  }
+
   return {
     connections: connections.map((connection) => {
       let dumpedConnection = {
--- a/test/context/directory/connections.test.js
+++ b/test/context/directory/connections.test.js
@@ -266,6 +266,69 @@ describe('#directory context connections', () => {
     expect(fs.existsSync(path.join(connectionsFolder, 'excludedConnection.json'))).to.equal(true);
   });
 
+  it('should only dump included connections', async () => {
+    const dir = path.join(testDataDir, 'directory', 'connectionsDumpInclude');
+    cleanThenMkdir(dir);
+    const context = new Context({ AUTH0_INPUT_FILE: dir }, mockMgmtClient());
+
+    context.assets.connections = [
+      { name: 'includedConnection', strategy: 'waad' },
+      { name: 'unmanagedConnection', strategy: 'samlp' },
+    ];
+    context.assets.include = { connections: ['includedConnection'] };
+
+    await handler.dump(context);
+    const connectionsFolder = path.join(dir, constants.CONNECTIONS_DIRECTORY);
+
+    expect(fs.existsSync(path.join(connectionsFolder, 'includedConnection.json'))).to.equal(true);
+    expect(fs.existsSync(path.join(connectionsFolder, 'unmanagedConnection.json'))).to.equal(false);
+  });
+
+  it('should preserve pre-existing files for connections outside the include list on re-dump', async () => {
+    const dir = path.join(testDataDir, 'directory', 'connectionsDumpInclude');
+    cleanThenMkdir(dir);
+
+    // Simulate a prior export that wrote a connection now outside the include list
+    const connectionsFolder = path.join(dir, constants.CONNECTIONS_DIRECTORY);
+    fs.ensureDirSync(connectionsFolder);
+    fs.writeFileSync(
+      path.join(connectionsFolder, 'unmanagedConnection.json'),
+      '{"name":"unmanagedConnection"}'
+    );
+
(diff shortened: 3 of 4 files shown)


jithinzac:
Looks like E2E failures are due to missing env variables

harshithRai (dismissed):

harshithRai:
@jithinzac changes look good! Can you please make sure your commits are signed to make it merge ready

jithinzac:
> @jithinzac changes look good! Can you please make sure your commits are signed to make it merge ready

Sorry for the delay. The commit is now signed.

harshithRai (approved):

Issue #1432, opened 2026-07-23
Deprecated external paths for actions throwing warnings for relative config paths


Nanovox:
### Checklist

- [x] I have looked into the [README](https://github.com/auth0/auth0-deploy-cli#readme) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://github.com/auth0/auth0-deploy-cli#documentation) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0-deploy-cli/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/auth0/auth0-deploy-cli/releases/latest) of this tool and the issue still persists.
- [x] I have searched the [Auth0 Community](https://community.auth0.com/tag/auth0-deploy-cli) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

The changes made in #1388 and subsequent warning message in #1392 are throwing warnings for action code mapped to a relative path.

We're using directory format for our configuration and the auth0-deploy-cli to deploy the configuration by executing the following:

`a0deploy import --input_file ./tenant/config`

We have several actions that are laid out in paths as follows:
- `./tenant/config/actions/action_one.json`
- `./tenant/config/actions/action_one/do_action_one.js`

The configuration for the action looks like this:
```
{
   "name": "Action One",
   "code": "actions/action_one/do_action_one.js"
   ...
}
```

Performing the deploy produces warnings as follows despite the relative paths:

> warn: Support for absolute paths and paths outside the config root will be deprecated in a future version to improve the security of the tool. Please update your configuration to use paths relative to the config directory. Current absolute path used: ["tenant/config/actions/action_one/do_add_one.js"]

It seems that the path is being resolved to an external path before being evaluated as a relative config path.

We've tried several things to get this to fall into the non-warned code path, but it seems to behave one of these three ways:

(1) It finds the code file and produce the warning -- `action.code` set to `actions/action_one/do_action_one.js`

> warn: Support for absolute paths...

(2) It doesn't fine the path and dies -- `action.code` set to `action_one/do_action_one.js`

> error: Unable to load file /workingdir/tenant/config/action_one/do_action_one.js due to Error: ENOENT: no such file or directory, access '/workingdir/tenant/config/action_one/do_action_one.js'

[...]

PR #1449, opened 2026-08-04
fix: address path traversal vulnerability and false warnings in config file handlers

ankita10119:
#### Background and motivation

This PR addresses two related problems reported against the 8.x release line:

**1. Security vulnerability (SEC-23449) - path traversal**
Resource configurations (`actions`, `actionModules`, `rules`, `hooks`, `databases`) reference external code files by path. A malicious or misconfigured path such as `../../sensitive-file.js` could cause the CLI to load a file from outside the intended config directory. This is a path traversal vulnerability reported by the security team.

**2. False deprecation warnings on legitimate relative paths (issue #1432)**
Users reported that the CLI was incorrectly emitting deprecation warnings for valid relative paths such as `./actions/action-one/code.js`. Investigation confirmed the root cause: the actions and actionModules handlers were using a fragile regex to pre-process paths before passing them to `loadFile()`, and `loadFile()` itself had a silent `isFile` fallback (`isFile(inSubfolder) ? inSubfolder : inRoot`) that behaved unpredictably depending on whether the file existed at the expected location.

These two problems share the same root cause: inconsistent and fragile path resolution logic across handlers.

  ---

#### Root cause

`databases.ts` already had the correct approach, it used `path.resolve + startsWith(configRoot)` directly in the handler. However, `actions.ts` and `actionModules.ts` used a regex to strip path prefixes before delegating to `loadFile()`:

 ```typescript
  // Old approach — fragile regex stripping caused false warnings
  const unixPath = action.code.replace(/[\\/]+/g, '/').replace(/^([a-zA-Z]+:|\.\/)/, '');
  action.code = context.loadFile(unixPath, actionFolder);
```

The regex incorrectly stripped the `./` prefix, causing `loadFile()` to resolve against the wrong base path and trigger false warnings for valid paths.

`rules.ts` and `hooks.ts` passed paths directly to `loadFile()` without any pre-processing, relying on the `isFile` fallback which could silently load the wrong file.

  ---

Edge cases identified and addressed

  1. Windows backslash paths - action.code can contain actions\\code.js on Windows. Fixed by normalizing backslashes with .replace(/\\/g, '/') before calling path.resolve.
  2. Action code path relative to config root - Action JSON stores the code path relative to the config root (e.g. "./actions/action-one/code.js"). The old regex stripped `./`, causing misresolution. Fixed by resolving directly from context.filePath using path.resolve.
  3. Rule and hook script paths relative to their subfolder - Rule JSON stores script paths relative to the rules/ subfolder (e.g. "somerule.js"), not the config root. Fixed by resolving from `configRoot/rules/` in `rules.ts` and `configRoot/hooks/` in `hooks.ts`.
  4. Absolute paths passed to `loadFile()` - On Unix, `path.join(base, absolutePath)` silently discards the base. Fixed by using `path.resolve()` throughout, which handles absolute paths correctly.
  5. Silent isFile fallback - The old `loadFile()` tried inSubfolder, and if the file didn't exist there, silently fell back to inRoot with no error or warning, potentially loading the wrong file. Removed entirely, path is now resolved deterministically with no fallback.
  6. Monorepo setups with shared code outside config root - Some teams store shared action code in a directory above the config root. A hard block would break these legitimate setups. Addressed with the `AUTH0_ALLOW_EXTERNAL_CODE_PATHS` escape hatch (see below).

  ---
What changed

  - `actions.ts` / `actionModules.ts`: Replaced regex pre-processing + `loadFile()` delegation with an explicit `path.resolve` + startsWith check, directly calling `loadFileAndReplaceKeywords`. Now consistent with `databases.ts`.
  - `rules.ts` / `hooks.ts`: Replaced `loadFile()` delegation with the same inline `path.resolve` + `startsWith` check.
  - `directory/index.ts` (`loadFile()`): Removed the `isFile` fallback. Emits a deprecation warning when path resolves outside config root. This will become a hard error in `v9.0.0`.
  - `yaml/index.ts` (`loadFile()`): Same fix - emits a deprecation warning when path resolves outside config root.
  - `databases.ts`: Detection logic was already correct; updated to support `AUTH0_ALLOW_EXTERNAL_CODE_PATHS `consistently with other handlers.
  - `types.ts`: Added `AUTH0_ALLOW_EXTERNAL_CODE_PATHS?: boolean` to the Config type.
  - `docs/configuring-the-deploy-cli.md`: Added documentation for `AUTH0_ALLOW_EXTERNAL_CODE_PATHS` including the monorepo use case, directory structure example, upgrade guidance, and security notice.

### 📚 References

Unit tests have been added or updated for all affected handlers:

  - `test/context/directory/actions.test.js`: Added tests for valid relative paths (no warning), path traversal warning, escape hatch (`AUTH0_ALLOW_EXTERNAL_CODE_PATHS`), and Windows-style backslash paths.
  - `test/context/directory/rules.test.js`: Added test asserting a deprecation warning is emitted when script path resolves outside config root.
  - `test/context/directory/hooks.test.js`: Added test asserting a deprecation warning is emitted when script path resolves outside config root.

All 25 tests in the above files pass.

### 🔬 Testing

Manual testing:
  - Point an action JSON at a path using `../../ traversal` - verify the deprecation warning fires and the deploy continues (`8.x `behavior).
  - Point an action JSON at a valid relative path (`./actions/action-one/code.js`) - verify it deploys with no warning.
  - Set `AUTH0_ALLOW_EXTERNAL_CODE_PATHS: true` with an external path - verify it loads successfully with no warning.

### 📝 Checklist


harshithRai (dismissed):
LGTM, provided v9 (#1448) lands the hard-error enforcement as I see it does there, since this only warns.

harshithRai (approved):

PR #1453, opened 2026-08-05
fix: exclude org-scoped roles from tenant export by filtering type='tenant'

harshithRai:
### 🔧 Changes

Excludes organization-scoped roles from the tenant roles export/import flow.

When the `api2_org_level_roles_ea` feature flag is enabled on a tenant, `GET /api/v2/roles` starts returning organization-scoped roles (those with `type: "organization"`) mixed in with tenant-level roles. The deploy CLI treats every listed role as tenant configuration, so without this change it would:

- pull org-scoped roles into the tenant baseline (`tenant.yaml` / `roles/*.json`)
- report spurious diffs for them on every run
- flag them for deletion when `AUTH0_ALLOW_DELETE` is enabled, since they are not present in the local config files

The roles handler now passes `type: 'tenant'` to `roles.list` so only tenant-level roles are managed. Org-scoped roles are managed within an organization context and are intentionally out of scope here.

The endpoint stays on offset pagination. Although the API spec moves `/roles` to checkpoint pagination, node-auth0 deliberately kept offset pagination for `/roles` for backward compatibility (checkpoint is deferred to a future major SDK), so `paginate: true` is retained rather than switching to checkpoint.

Bumps `auth0` to `^6.2.0`, which adds the `type` filter parameter and includes `type` on the `Role` response.

**Shape change:** exported roles now include a `type` field (returned by the 6.2.0 `Role` response). This is additive and backward compatible on import. Role IDs are stripped on export as before, so no `id` is written.

YAML (`tenant.yaml`):

```yaml
roles:
  - name: Test Role
    description: Role to reproduce the error
    permissions: []
    type: tenant   # new: present on exported roles
```

JSON(`roles/Test Role.json`):

```json
{
  "name": "Test Role",
  "description": "Role to reproduce the error",
  "type": "tenant",
  "permissions": []
}
```

#### Migration note

Configs exported while the `api2_org_level_roles_ea` flag was on may contain organization-scoped roles. After upgrading, these are no longer read from the tenant, so they appear local-only and would be recreated as tenant roles on deploy. Affected roles should be removed from local config after a re-export.

### 🔬 Testing

Unit tests: `test/tools/auth0/handlers/roles.tests.js` asserts `roles.list` is called with `type: 'tenant'`.

Manual end-to-end against a live tenant with `api2_org_level_roles_ea` enabled:

1. Created an org-scoped role via the Management API (`POST /api/v2/roles` with `type: "organization"` and an `owner_id`).
2. Confirmed at the API level that `GET /api/v2/roles` returns it (3 roles) while `GET /api/v2/roles?type=tenant` does not (2 roles).
3. Ran `export` in both YAML and directory formats: the org-scoped role was excluded from the output; only tenant roles were written, each with a `type` field and no `id`.
4. Ran `import --dry-run` with the org role present on the tenant: the proposed-changes plan was empty, confirming the org role is not pulled into the baseline and not flagged for deletion (even with `AUTH0_ALLOW_DELETE` enabled).
5. Deleted the temporary org-scoped role to restore tenant state.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/roles.ts
+++ b/src/tools/auth0/handlers/roles.ts
@@ -149,9 +149,12 @@ export default class RolesHandler extends DefaultHandler {
     }
 
     try {
+      // Only manage tenant-level roles; exclude org-scoped roles surfaced when
+      // the api2_org_level_roles_ea flag is enabled.
       const roles = await paginate<Role>(this.client.roles.list, {
         paginate: true,
         include_totals: true,
+        type: 'tenant',
       });
 
       for (let index = 0; index < roles.length; index++) {
--- a/src/tools/auth0/handlers/roles.ts
+++ b/src/tools/auth0/handlers/roles.ts
@@ -14,6 +14,7 @@ export const schema = {
       name: { type: 'string' },
       id: { type: 'string' },
       description: { type: 'string' },
+      type: { type: 'string', enum: ['tenant', 'organization'] },
       permissions: {
         type: 'array',
         items: {
@@ -110,6 +111,7 @@ export default class RolesHandler extends DefaultHandler {
 
     delete data.permissions;
     delete data.id;
+    delete data.type; // immutable; set at creation only, never send on update
 
     await this.client.roles.update(params.id, data);
 
--- a/src/tools/auth0/handlers/roles.ts
+++ b/src/tools/auth0/handlers/roles.ts
@@ -45,8 +45,8 @@ export default class RolesHandler extends DefaultHandler {
   async createRole(data): Promise<Asset> {
     const role = { ...data };
     delete role.permissions;
-    // `type` is a read-only field returned by the roles GET endpoint but rejected
-    // by the roles create/update endpoints. Strip it to keep exports round-trippable.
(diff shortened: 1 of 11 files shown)


src/tools/auth0/handlers/roles.ts
```diff
@@ -149,9 +151,12 @@ export default class RolesHandler extends DefaultHandler {
     }
 
     try {
+      // Only manage tenant-level roles; exclude org-scoped roles surfaced when
+      // the api2_org_level_roles_ea flag is enabled.
```
kushalshit27, line 160: What happens if for those tenents where the `api2_org_level_roles_ea`  flag is not enabled?

harshithRai: Nothing changes for them. `type` is a standard filter on `GET /api/v2/roles`, and on a flag-off tenant all roles are tenant-level anyway, so `type=tenant` returns the exact same set as no filter. On flag-on tenants it just excludes the org-scoped roles we never managed.

ankita10119 (commented):
Could you please add a test case for `databases.ts `changes

ankita10119 (dismissed):

ankita10119 (approved):

PR #1454, opened 2026-08-06
feat: add org-to-app entitlement (EA) support for organizations

ankita10119:
### 🔧 Changes

Add `is_app_entitlement_active` and org-client associations (clients) support for organizations (Org-to-App Entitlement EA).                                                       
                                                                                                                                                                                   
What changed:                          
  - `is_app_entitlement_active` boolean is now exported/deployed on all organizations when the `org_to_app_entitlement_enabled` feature flag is enabled on the tenant                  
  - New `clients` sub-resource on organizations: manages which client applications org members are entitled to use for login via `use_for_member_access  `                             
  - On export, `client_id` values are resolved to client names for portability across environments; on deploy, names are resolved back to IDs                                        
  - Create/delete operations are batched in chunks of 10 per API limits
  - Gracefully handles tenants without the feature flag enabled (403 → skips silently)
  - Requires `read:organization_clients`, `create:organization_clients`, `update:organization_clients`, `delete:organization_clients` scopes on the M2M application

### 📚 References


### 🔬 Testing

- Unit tests added for: export with client name resolution, create with org-clients, update (add/remove/update), remove all org-clients, and graceful 403 handling
  - Tested end-to-end against a real tenant with the `org_to_app_entitlement_enabled` FF enabled: export → modify org → deploy → re-export confirmed the association was applied correctly

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/organizations.ts
+++ b/src/tools/auth0/handlers/organizations.ts
@@ -505,9 +505,20 @@ export default class OrganizationsHandler extends DefaultHandler {
     }
 
     if (orgClientsToRemove.length > 0) {
-      await this.deleteOrganizationClients(params.id, orgClientsToRemove).catch((err) => {
-        throw new Error(`Problem removing org clients for organization ${params.id}\n${err}`);
-      });
+      if (
+        this.config('AUTH0_ALLOW_DELETE') === 'true' ||
+        this.config('AUTH0_ALLOW_DELETE') === true
+      ) {
+        await this.deleteOrganizationClients(params.id, orgClientsToRemove).catch((err) => {
+          throw new Error(`Problem removing org clients for organization ${params.id}\n${err}`);
+        });
+      } else {
+        log.warn(
+          `Detected the following organization client associations should be removed. Doing so may be destructive.\nYou can enable deletes by setting 'AUTH0_ALLOW_DELETE' to true in the config\n${orgClientsToRemove.join(
+            '\n'
+          )}`
+        );
+      }
     }
 
     await Promise.all(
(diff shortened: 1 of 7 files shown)


src/tools/auth0/handlers/organizations.ts
```diff
@@ -423,6 +463,67 @@ export default class OrganizationsHandler extends DefaultHandler {
...
+      await this.createOrganizationClients(params.id, orgClientsToAdd).catch((err) => {
+        throw new Error(`Problem adding org clients for organization ${params.id}\n${err}`);
+      });
+    }
+
+    if (orgClientsToRemove.length > 0) {
+      await this.deleteOrganizationClients(params.id, orgClientsToRemove).catch((err) => {
+        throw new Error(`Problem removing org clients for organization ${params.id}\n${err}`);
+      });
+    }
```
harshithRai, line 522: Should this match the `discovery_domains` pattern? That one gates removals behind AUTH0_ALLOW_DELETE and warns instead of deleting ([organizations.ts:406](https://github.com/auth0/auth0-deploy-cli/blob/e245f167723acc98b5961aa4f2c926119b2bfac4/src/tools/auth0/handlers/organizations.ts#L406)), whereas here clients removes unconditionally. A stale config with clients: [] would silently wipe out associations set up outside the CLI.

harshithRai:
LGTM! Just had the one open-ended question above

src/tools/auth0/handlers/organizations.ts
ankita10119: Good catch! You're right actually, keeping the `AUTH0_ALLOW_DELETE` guard is consistent with the broader pattern in the codebase where most sub-resource deletions (`clientAuthCredentials`, `SCIM` config, `discovery_domains`, directory provisioning configs etc.) are gated behind it. Org `connections` and `clientGrants` being unconditional look more like oversights than intentional design

Addressed and updated to match the pattern

harshithRai (approved):

PR #1457, opened 2026-08-10
fix: create phone templates on import instead of skipping when they have no ID

harshithRai:
### 🔧 Changes

Fixes phone notification template import failing against a **newly created tenant**.

On a fresh tenant, the Management API returns the default phone templates **without an `id`** until they are explicitly created. `calcChanges` matches assets by `type` and routes them all to `update`, but the update (PATCH) endpoint requires a template `id`. The result was every template being skipped:

~~~
warn: Skipping update for phone template type 'otp_verify' as unable to find existing template ID
~~~

`PhoneTemplatesHandler.processChanges` now:

- Splits the calculated `update` set into templates whose existing counterpart **has an `id`** (→ PATCH) vs. **has no `id` yet** (→ POST/create), so new-tenant templates are created instead of skipped.
- Adds a **409-conflict fallback** in `createPhoneTemplate`: if the template already exists (created between our `list` and `create`, or its `id` wasn't surfaced by `list`), we re-fetch the list to pick up the `id` and fall back to an update.
- Strips read-only create-only fields (`channel`, `customizable`, `tenant`) from the create payload via `stripFields(template, this.stripCreateFields)`, since `processChanges` is overridden here and doesn't go through the base handler's stripping.
- Omits `content.from` from the create/update payload when it is blank. Fresh tenants export `from: ''`, but the API rejects an empty string (`minimum 1` char), so it must be dropped rather than sent.

### 🔬 Testing

Unit tests added in `test/tools/auth0/handlers/phoneTemplates.test.ts` cover:

- Creating a template when the existing template has no `id` (new-tenant case) - POST, not PATCH.
- Omitting empty `content.from` from both create and update payloads.
- Falling back to update when create returns a 409 conflict (re-fetches the list to pick up the `id`).
- Stripping read-only fields (`channel`, `customizable`, `tenant`) from the create payload.
- Existing create / update / delete / no-op paths still pass (14 tests total).

Manually verified end-to-end against a freshly created tenant: export (4 default templates, all `from: ''`) → edit a body → import (all 4 created) → re-import (all 4 updated, idempotent) → re-export confirms the edit persisted.

### 📝 Checklist


src/tools/auth0/handlers/phoneTemplates.ts
```diff
@@ -97,18 +98,56 @@ export default class PhoneTemplatesHandler extends DefaultHandler {
...
   async createPhoneTemplate(template): Promise<Asset> {
-    const created = await this.client.branding.phone.templates.create(template);
-    return created;
+    // The create endpoint only accepts type/disabled/content; strip read-only
+    // fields (channel, customizable, tenant) that the API would reject.
+    const createPayload = stripFields(template, this.stripCreateFields);
+    // `content.from` is optional but the API rejects an empty string. Fresh
+    // tenants export it as '', so drop it when blank to let the create succeed.
+    if (createPayload.content && !createPayload.content.from) {
+      delete createPayload.content.from;
```
ankita10119, line 136: `stripFields` does a shallow copy of the top-level object, so `createPayload.content` and `template.content` point to the same object reference. The `delete createPayload.content.from` on the next line therefore mutates the original `template.content` in place.

This is incidentally harmless right now, the 409 fallback calls `updatePhoneTemplate(template)`, which also skips a blank `from`, so the side effect is correct by accident. But if the fallback path or `updatePhoneTemplate`'s contract changes, this could silently produce wrong behaviour

scripts/verify-phone-templates-fix.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
```
ankita10119: Do we need this file?

ankita10119 (changes requested):

src/tools/auth0/handlers/phoneTemplates.ts
harshithRai: Fixed in commit 00052f4.. had missed this commit

scripts/verify-phone-templates-fix.sh
harshithRai: Fixed in commit https://github.com/auth0/auth0-deploy-cli/commit/00052f49e4c49413615e4d4116c3600024240f02.. had missed this commit

ankita10119 (approved):

PR #1465, opened 2026-08-19
fix: preserve idpinitiated.enabled:false on SAML connection export instead of emitting undefined_clientId

harshithRai:
### 🔧 Changes

Fixes a serialization bug where exporting a SAML (samlp) connection with IDP-initiated login disabled produced a malformed `idpinitiated` block. Previously the exporter unconditionally injected a `client_id`, and when none existed the helper substituted the literal string `undefined_clientId`. On import this present block was treated as enabled, silently flipping IDP-initiated login ON in the target tenant.

- `getFormattedOptions` in `src/utils.ts` (export path, ID to name) now only adds `client_id` when one is actually present; otherwise the `idpinitiated` block passes through untouched, preserving `enabled: false`.
- `getFormattedOptions` in `src/tools/auth0/handlers/connections.ts` (diff/import path, name to ID) receives the same guard so both directions stay symmetric. A third resolver in `src/tools/calculateDryRunChanges.ts` was reviewed and is already guarded by `?.client_id`, so it needs no change.

Note on scope: the guard omits `client_id` whenever it is falsy (`enabled: false`, empty string, null, or absent), not only the disabled case. This is the intended, more-correct behavior, but the output change is slightly wider than "disabled connections" alone.

**Exported config shape change (both YAML and JSON/directory formats):**

For a connection with IDP-initiated login disabled, the export now emits:

```yaml
idpinitiated:
  enabled: false
```

```json
"idpinitiated": {
  "enabled": false
}
```

instead of the previous malformed output:

```yaml
idpinitiated:
  client_id: undefined_clientId
```

```json
"idpinitiated": {
  "client_id": "undefined_clientId"
}
```

Connections that have IDP-initiated login enabled are unaffected and continue to export their resolved `client_id`.

### 🔬 Testing

- Added a disabled IDP-initiated connection to the YAML dump test, asserting `enabled: false` round-trips without a bogus `client_id`. This covers both the YAML and directory export paths, since both share `utils.getFormattedOptions`.
- Added a focused unit test on the handler's `getFormattedOptions` (the import/diff name-to-ID path) verifying no `client_id` is re-injected when a disabled `{ enabled: false }` block is processed - directly pinning the reported "login flipped ON" symptom.
- Full connections, dry-run, and YAML/directory context suites pass.

Verified end-to-end against a real tenant:

- Created a samlp connection with `idpinitiated.enabled: false` via import.
- Pre-fix build reproduced the bug: the export emitted `enabled: false` **and** a stray `client_id: undefined_clientId`.
- Fixed build exported the block as `enabled: false` only, with no `undefined_clientId` placeholder, so the disabled state round-trips correctly.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/connections.ts
+++ b/src/tools/auth0/handlers/connections.ts
@@ -448,12 +448,15 @@ export default class ConnectionsHandler extends DefaultAPIHandler {
 
   getFormattedOptions(connection, clients) {
     try {
+      const { idpinitiated } = connection.options;
       return {
         options: {
           ...connection.options,
           idpinitiated: {
-            ...connection.options.idpinitiated,
-            client_id: convertClientNameToId(connection.options.idpinitiated.client_id, clients),
+            ...idpinitiated,
+            ...(idpinitiated.client_id && {
+              client_id: convertClientNameToId(idpinitiated.client_id, clients),
+            }),
           },
         },
       };
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -334,12 +334,15 @@ export const decodeBase64ToCertString = (base64Cert: string) => {
 // Format connection options by converting client IDs to client names for SAML connections
 export const getFormattedOptions = (connection, clients) => {
   try {
+    const { idpinitiated } = connection.options;
     return {
       options: {
         ...connection.options,
         idpinitiated: {
-          ...connection.options.idpinitiated,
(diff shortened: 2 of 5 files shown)


ankita10119 (dismissed):
Can you add a parallel test case in `test/context/directory/connections.test.js` for a disabled idpinitiated block to match the YAML coverage?

ankita10119 (approved):

PR #1471, opened 2026-08-24
feat: support third_party_client_access and cross_app_access_resource_app

harshithRai:
### 🔧 Changes

Adds support for two new Early Access properties introduced in node-auth0 6.3.0. Bumps the `auth0` dependency to `^6.3.0`.

**1. `clients` -> `my_organization_configuration.third_party_client_access`**

Controls whether third-party clients can access the organization via the My Organization API. Shape: `{ default_value, allowed_values }` where `default_value` is `block` and `allowed_values` is an array of `allow` / `block`.

**2. `connectionProfiles` -> `cross_app_access_resource_app`**

Controls whether organization admins may enable Cross App Access (XAA) on their Identity Providers. Shape: `{ status: { default_value, allowed_values } }` where `default_value` is `enabled` / `disabled` and `allowed_values` is an array of `enabled` / `disabled`.

Both properties are optional and are gated behind tenant feature flags on the Management API side.

#### YAML examples

```yaml
clients:
  - name: "My Enterprise App"
    my_organization_configuration:
      allowed_strategies:
        - okta
      connection_deletion_behavior: allow_if_empty
      third_party_client_access:
        default_value: block
        allowed_values:
          - allow
          - block

connectionProfiles:
  - name: "Enterprise SSO Profile"
    enabled_features:
      - scim
    cross_app_access_resource_app:
      status:
        default_value: enabled
        allowed_values:
          - enabled
          - disabled
```

#### JSON (directory format) examples

`clients/My Enterprise App.json`

```json
{
  "name": "My Enterprise App",
  "my_organization_configuration": {
    "allowed_strategies": ["okta"],
    "connection_deletion_behavior": "allow_if_empty",
    "third_party_client_access": {
      "default_value": "block",
      "allowed_values": ["allow", "block"]
    }
  }
}
```

`connection-profiles/Enterprise SSO Profile.json`

```json
{
  "name": "Enterprise SSO Profile",
  "enabled_features": ["scim"],
  "cross_app_access_resource_app": {
    "status": {
      "default_value": "enabled",
      "allowed_values": ["enabled", "disabled"]
    }
  }
}
```

### 🔬 Testing

- Added unit tests covering AJV schema validation for both new properties (valid payloads pass, missing required sub-fields fail).
- Verified end to end against a live tenant: `import` correctly serializes and sends both new fields to the Management API. On a tenant without the feature flags enabled, the API returns the expected `403 operation_not_supported` responses, confirming the fields reach the API rather than being stripped:
  - `third_party_client_access`: "The account is not allowed to set third_party_client_access"
  - `cross_app_access_resource_app`: "The account is not enabled to set cross_app_access_resource_app"
- Full round-trip create + export requires these tenant feature flags: `organizations_third_party_client_support`, `my_orgs_cross_app_access_resource_app`, `cross_app_access_resource_application_beta`.

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/connectionProfiles.ts
+++ b/src/tools/auth0/handlers/connectionProfiles.ts
@@ -66,6 +66,8 @@ export const schema = {
                 description: 'The allowed Cross App Access resource app status values.',
               },
             },
+            // allowed_values is optional per the SDK type (unlike the client's
+            // third_party_client_access, which requires both fields).
             required: ['default_value'],
           },
         },
--- a/src/tools/auth0/handlers/connectionProfiles.ts
+++ b/src/tools/auth0/handlers/connectionProfiles.ts
@@ -39,6 +39,38 @@ export const schema = {
       connection_config: {
         type: ['object', 'null'],
       },
+      cross_app_access_resource_app: {
+        type: 'object',
+        description:
+          'Controls whether organization admins may enable Cross App Access (XAA) on their Identity Providers.',
+        properties: {
+          status: {
+            type: 'object',
+            properties: {
+              default_value: {
+                type: 'string',
+                enum: Object.values(
+                  Management.ConnectionProfileCrossAppAccessResourceAppStatusDefaultValueEnum
+                ),
+                description: 'The default Cross App Access resource app status.',
+              },
+              allowed_values: {
+                type: 'array',
+                items: {
+                  type: 'string',
+                  enum: Object.values(
+                    Management.ConnectionProfileCrossAppAccessResourceAppStatusValueEnum
+                  ),
+                },
+                uniqueItems: true,
+                description: 'The allowed Cross App Access resource app status values.',
+              },
+            },
+            required: ['default_value'],
+          },
+        },
+        required: ['status'],
+      },
       strategy_overrides: {
         type: ['object', 'null'],
         properties: {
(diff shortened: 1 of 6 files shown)


src/tools/auth0/handlers/connectionProfiles.ts
```diff
@@ -39,6 +39,38 @@ export const schema = {
...
+                  type: 'string',
+                  enum: Object.values(
+                    Management.ConnectionProfileCrossAppAccessResourceAppStatusValueEnum
+                  ),
+                },
+                uniqueItems: true,
+                description: 'The allowed Cross App Access resource app status values.',
+              },
+            },
+            required: ['default_value'],
```
ankita10119, line 71: `third_party_client_access` in `clients.ts` marks both `default_value` and `allowed_values` as required, but `cross_app_access_resource_app.status` here only requires `default_value` , making `allowed_values` optional. Is this difference intentional based on what each Management API endpoint actually accepts? If the client API rejects a payload without `allowed_values` but the connection profile API accepts it, that's fine and worth a comment here. But if both APIs behave the same way, these schemas should be consistent with each other

harshithRai: Yes, this is intentional.. matches the node-auth0 SDK types. `ClientMyOrganizationThirdPartyClientAccessConfiguration` requires both fields, but `ConnectionProfileCrossAppAccessResourceAppStatus` marks `allowed_values` optional. So the client API rejects a payload without `allowed_values` while the connection-profile API accepts it. Added a comment to document it.

ankita10119 (approved):

PR #1479, opened 2026-08-28
fix: prevent false enabled_clients diff in database dry-run

harshithRai:
### 🔧 Changes

Fixes a false positive in `--dry-run` mode where database connections with `enabled_clients` were incorrectly reported as changed.

In the database dry-run path (`dryRunChanges`), the local connection had its `enabled_clients` populated, but the remote connection it was compared against was fetched from `connections.list`, which no longer returns `enabled_clients` inline. The field now lives behind the dedicated enabled-clients endpoint. Because the remote object lacked the key entirely, the diff reported a false difference:

````
Key [<connection>.enabled_clients] found in 'localObj' but not in 'remoteObj'.
````

The fix enriches the remote connections with their `enabled_clients` via `getConnectionEnabledClients` before diffing, mirroring what the real import path already does in `getType()`. The change is scoped entirely to `dryRunChanges`, so real import and export behavior is untouched.

### 📚 References

- DXCDT-2281
- Related dry-run reporting fixes: #1462

### 🔬 Testing

Added a regression test in `test/tools/auth0/handlers/dryRun.tests.ts` that fails on the old code (reports a spurious update) and passes with the fix.

Verified end to end against live tenants: exported a tenant to a directory, then ran a dry-run against the identical export. On tenants whose `connections.list` omits `enabled_clients` inline, the old code produced the false positive and the fix removes it. Confirmed the real import and export paths are unaffected.

To reproduce manually:

````
npx a0deploy export -c=config.json --format=directory --output_folder=local
npx a0deploy import -c=config.json --input_file=local --dry-run --debug
````

### 📝 Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ Makefile
 .github/skills/*
 package-lock.json
 .github/hooks/*
+.DS_Store
--- a/src/tools/auth0/handlers/databases.ts
+++ b/src/tools/auth0/handlers/databases.ts
@@ -605,6 +605,22 @@ export default class DatabaseHandler extends DefaultAPIHandler {
       }),
     ]);
 
+    // `enabled_clients` is no longer returned by connections.list; it must be fetched from the
+    // dedicated enabled-clients endpoint. Without this enrichment the remote connection lacks the
+    // field entirely, so the dry-run diff reports a false "found in localObj but not in remoteObj".
+    const existingWithEnabledClients = await Promise.all(
+      existingDatabasesConnections.map(async (con) => {
+        if (!con?.id) return con;
+
+        const enabledClients = await getConnectionEnabledClients(this.client, con.id);
+        if (enabledClients && enabledClients.length) {
+          return { ...con, enabled_clients: enabledClients };
+        }
+
+        return con;
+      })
+    );
+
     const formatted = databases.map((db) => {
       const { options, ...rest } = db;
       const formattedOptions = this.getFormattedOptions(options, actions);
@@ -614,7 +630,7 @@ export default class DatabaseHandler extends DefaultAPIHandler {
         formattedDb.enabled_clients = getEnabledClients(
           assets,
           db,
-          existingDatabasesConnections,
+          existingWithEnabledClients,
           clients
         );
       }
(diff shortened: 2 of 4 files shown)


ankita10119 (commented):
Shouldn't `calcChanges` also enrich with `enabled_clients` the same way `dryRunChanges` does after this fix? If `connections.list` no longer returns `enabled_clients` inline, then `getEnabledClients` in `calcChanges` would see no `enabled_clients` on the remote connection, which means excluded clients wouldn't be preserved and could get silently disabled on a real import. Am I reading this correctly?

ankita10119:
`getType()`, `calcChanges`, and `dryRunChanges` all independently call `connections.list` and enrich (or forget to enrich) `enabled_clients` separately. Would it be worth extracting the enrichment into a shared helper so these paths can't drift out of sync? I understand this is out of scope, just flagging it while we're touching this workflow

harshithRai:
> Shouldn't `calcChanges` also enrich with `enabled_clients` the same way `dryRunChanges` does after this fix? If `connections.list` no longer returns `enabled_clients` inline, then `getEnabledClients` in `calcChanges` would see no `enabled_clients` on the remote connection, which means excluded clients wouldn't be preserved and could get silently disabled on a real import. Am I reading this correctly?

Yes that understanding is correct. And it's a valid catch too. Fixed both in the last 2 commits.

`calcChanges` now enriches too, so excluded-but-enabled clients are preserved on a real import instead of getting silently disabled. I confirmed it live on a test tenant: before the fix the excluded client was disabled, after it stays enabled.

> `getType()`, `calcChanges`, and `dryRunChanges` all independently call `connections.list` and enrich (or forget to enrich) `enabled_clients` separately. Would it be worth extracting the enrichment into a shared helper so these paths can't drift out of sync? I understand this is out of scope, just flagging it while we're touching this workflow

I also pulled the enrichment into a single shared helper (`enrichConnectionsWithEnabledClients`) that `getType`, `calcChanges`, and `dryRunChanges` all use, so the paths can't drift. Small bonus: calcChanges reuses `getType`'s cached result rather than re-fetching, which also dropped a duplicate connections.list call.

ankita10119 (approved):

PR #1463, opened 2026-08-18
Guard undefined `handler.identifiers` in keywordPreservation.ts

TimeTravelerFromNow:
### 🔧 Changes

This rescues an error in _src/keywordPreservation.ts_ when `handler.identifiers` is `undefined`
https://github.com/auth0/auth0-deploy-cli/issues/1446

### 📚 References

Uses @rdroe 's suggested fix _Option B_ described in https://github.com/auth0/auth0-deploy-cli/issues/1446

> Option B — skip handlers without identifiers:
> ```typescript
> if (handler.identifiers === undefined) return acc;
> acc[handler.type] = handler.identifiers.flat();
> ```

### 🔬 Testing

With `AUTH0_PRESERVE_KEYWORDS: true` in config, perform an export for example like so:

```bash
npx auth0-deploy-cli export \
  --strip \
  --format directory \
  --output_folder ./${output_folder} \
  --config_file ./${output_folder}/${from_tenant}.json \
  --debug true
```
### 📝 Checklist

- [x] All new/changed/fixed functionality is covered by tests (or N/A)
- [N/A] I have added documentation for all new/changed functionality (or N/A)

PR #1485, opened 2026-09-04
fix: allow system resource servers other than the My Account API to be updated

### LINKED ISSUE #1484: resourceServers: system resource servers other than Auth0 My Account API fail on every update (Additional properties not allowed: name)
### Checklist

### Description

`ResourceServersHandler.updateResourceServer()` strips `name` from the update payload only for the Auth0 My Account API, and it matches on that API's literal name. Any *other* system resource server takes the full-update path, which sends `name`, and the Management API rejects it:

```
Payload validation error: 'Additional properties not allowed: name'
```

We hit this with the **Auth0 My Organization API** (`https://{domain}/my-org/`), which is a system resource server just like the My Account API.

Because `calculateChanges()` puts every asset it matches in the tenant into `update` regardless of whether anything actually changed, this happens on *every* run once the resource server exists. The first `import` succeeds (create strips `is_system`); the second one fails and aborts the whole `processChanges` stage for `resourceServers`, so the type is left half-applied.

The `is_system` half of the guard never fires:

```ts
// src/tools/auth0/handlers/resourceServers.ts
stripUpdateFields: ['identifier', 'client_id', 'is_system'],
...
if (update.is_system === true || update.name === 'Auth0 My Account API') {
```

`stripFields()` runs on the payload in `DefaultHandler.processChanges()` *before* `updateResourceServer()` is called, so `update.is_system` is always `undefined`. Only the hardcoded name comparison does anything, which is why the My Account API happens to work and nothing else does.

### Expectation

Any system resource server round-trips. `import` should be idempotent for `Auth0 My Organization API` in the same way it already is for `Auth0 My Account API`.

### Reproduction

Reproduces consistently, every time.

1. Given a tenant with the My Organization API activated (the **Activate** button under Applications > APIs, or `POST /api/v2/resource-servers` with identifier `https://{domain}/my-org/`), and this in the config directory:

   ```json
   // resource-servers/Auth0 My Organization API.json
   {
     "name": "Auth0 My Organization API",
     "identifier": "https://{domain}/my-org/"
   }
   ```

2. When you run `a0deploy import` twice — the first run creates the resource server if it is not there yet, the second run updates it

3. Then the second run fails:

   ```
   info: Updated [resourceServers]: {"name":"Auth0 My Account API","identifier":"https://{domain}/me/"}
   error: Problem running command import during stage processChanges when processing type resourceServers
   error: Problem updating resourceServers {"name":"Auth0 My Organization API","identifier":"https://{domain}/my-org/"}
   BadRequestError: BadRequestError
   Status code: 400
   Body: {
     "statusCode": 400,
     "error": "Bad Request",
     "message": "Payload validation error: 'Additional properties not allowed: name'.",
     "errorCode": "invalid_body"
   }
   ```

Putting a file for `Auth0 My Account API` next to it makes the contrast visible: in the same run that one updates fine and the My Organization API does not.

### Suggested fix

Read `is_system` off the existing resource server instead of off the stripped update payload, and drop the name comparison:

```ts
const existing = this.existing?.find((rs) => rs.id === id);
if (existing?.is_system === true) {
  ...
}
```

`getType()` already keeps `is_system` on what it caches in `this.existing`, and it runs before `processChanges()`. Happy to open a PR — I have one ready.

### Related

`getType()`'s allowlist for system resource servers keeps `authorization_policy` but not `subject_type_authorization`, while `updateResourceServer()` does send the latter. Minor and separate, but it means that field cannot round-trip through an export.

### Deploy CLI version

8.44.0 (also reproduced on 8.35.0)

### Node version

24.12.0

erdtsieck:
### 🔧 Changes

`updateResourceServer()` decides whether to send the restricted payload with this guard:

```ts
if (update.is_system === true || update.name === 'Auth0 My Account API') {
```

Neither half holds up:

- `is_system` is listed in `stripUpdateFields`, and `stripFields()` runs on the payload in
  `DefaultHandler.processChanges()` *before* `updateResourceServer()` is called. `update.is_system`
  is therefore always `undefined` and that branch is dead code.
- The fallback matches one hardcoded name, so the Auth0 My Account API is the only system resource
  server that works. Every other one — the Auth0 My Organization API (`https://{domain}/my-org/`) in
  our case — takes the full-update path, which includes `name`, and the Management API rejects it
  with `Payload validation error: 'Additional properties not allowed: name'`.

Since `calculateChanges()` puts every matched asset into `update` whether or not anything changed,
this makes `import` fail on every run after the first, and it aborts the whole `processChanges` stage
for `resourceServers`.

This reads `is_system` off the existing resource server instead, which is where it actually survives —
`getType()` keeps `is_system` in its allowlist when it sanitizes system resource servers, and it runs
before `processChanges()`. The hardcoded name comparison then has nothing left to do and is removed:
the My Account API is covered by `is_system` like any other system resource server.

No change in behaviour for the My Account API or for non-system resource servers.

### 📚 References

Fixes #1484

### 🔬 Testing

Added `should update a system resource server other than "Auth0 My Account API" without name`, which
mirrors the existing My Account API test with the My Organization API. On master it fails with
`expected 'Auth0 My Organization API' to equal undefined`, which is exactly the `name` that the
Management API rejects; with this change it passes.

The existing system resource server tests (sanitizing in `getType`, `authorization_policy` in the
update payload, `should update "Auth0 My Account API" without name and is_system`) all still pass, so
the removed name comparison is not load-bearing.

Verified end-to-end against a real tenant as well, with `AUTH0_INCLUDED_ONLY: ["resourceServers"]` and
a config directory holding both system APIs. On 8.44.0 the first `import` creates both and the second
one fails on the My Organization API; with this change both runs are clean.

### 📝 Checklist

- [x] I have added documentation for all new/changed functionality (or N/A) — N/A, bug fix with no
      public API or config surface change

### DIFF
--- a/src/tools/auth0/handlers/resourceServers.ts
+++ b/src/tools/auth0/handlers/resourceServers.ts
@@ -199,8 +199,13 @@ export default class ResourceServersHandler extends DefaultHandler {
     id: string,
     update: ResourceServer
   ): Promise<Management.UpdateResourceServerResponseContent> {
-    // Exclude name from update as it cannot be modified for system resource servers like Auth0 My Account API
-    if (update.is_system === true || update.name === 'Auth0 My Account API') {
+    // Exclude name from update as it cannot be modified for system resource servers like the
+    // Auth0 My Account API or the Auth0 My Organization API. `is_system` is listed in
+    // `stripUpdateFields`, so it has already been removed from `update` by the time this runs -
+    // read it off the existing resource server instead, which `getType()` retains it on.
+    const existing = this.existing?.find((resourceServer) => resourceServer.id === id);
+
+    if (existing?.is_system === true) {
       const updateFields: Management.UpdateResourceServerRequestContent = {
         token_lifetime: update.token_lifetime,
         proof_of_possession: update.proof_of_possession,
--- a/test/tools/auth0/handlers/resourceServers.tests.js
+++ b/test/tools/auth0/handlers/resourceServers.tests.js
@@ -964,5 +964,47 @@ describe('#resourceServers handler', () => {
       await stageFn.apply(handler, [data]);
       expect(updateCalled).to.equal(true);
     });
+
+    it('should update a system resource server other than "Auth0 My Account API" without name', async () => {
+      let updateCalled = false;
+      const existingResourceServer = {
+        id: 'rs_my_organization',
+        identifier: 'https://auth0.com/my-organization/my-org/',
+        name: 'Auth0 My Organization API',
+        is_system: true,
+      };
+
+      const auth0 = {
+        resourceServers: {
+          create: () => Promise.resolve({ data: [] }),
+          update: function (id, data) {
+            updateCalled = true;
+            expect(id).to.equal('rs_my_organization');
+            expect(data.name).to.equal(undefined);
+            expect(data.is_system).to.equal(undefined);
+            expect(data.token_lifetime).to.equal(600);
+            return Promise.resolve({ data });
+          },
+          delete: () => Promise.resolve({ data: [] }),
+          list: (params) => mockPagedData(params, 'resource_servers', [existingResourceServer]),
+        },
+        pool,
+      };
+
+      const handler = new resourceServers.default({ client: pageClient(auth0), config });
+      const stageFn = Object.getPrototypeOf(handler).processChanges;
+      const data = {
+        resourceServers: [
+          {
+            name: 'Auth0 My Organization API',
+            identifier: 'https://auth0.com/my-organization/my-org/',
+            token_lifetime: 600,
+          },
+        ],
+      };
+
+      await stageFn.apply(handler, [data]);
+      expect(updateCalled).to.equal(true);
+    });
   });
 });


erdtsieck:
A note on the red **E2E tests as Node module** check, since it is not caused by this change:

- The four failures are all `Problem updating clients {"name":"Deploy CLI", ...}` with `Nock: No match for request` on `PATCH /api/v2/clients/Vp0gMRF8...`. Nothing in them touches resource servers — the recordings only contain a `GET /api/v2/resource-servers`, no update, so `updateResourceServer()` is never reached in that suite.
- Running `AUTH0_HTTP_RECORDINGS=lockdown npm run test:e2e:node-module` locally on this branch and on an unmodified `master` gives byte-for-byte the same failures, including the same four.
- The job appears to fail on every fork PR (#1473, #1463, #1442) and pass on every PR pushed to an `auth0`-owned branch, which is what you would expect if the job needs a CircleCI context that fork builds do not receive.

`Unit tests with Node LTS`, `Unit tests with Node current`, TypeScript, lint and format are all green, and those cover the added regression test.

Happy to rebase or adjust anything if you would rather see this shaped differently.

harshithRai:
Really appreciate this, the fix and the tests both look great and everything's passing on my end. Thank you for the thorough write-up too.

Just one thing before I can merge: the commits need to be signed and verified for it to be merge-ready. Could you re-sign and force-push? It'd be good to bring the branch up to date with master while you're at it (we're on node-auth0 v7 now, applies cleanly).

Once the commits show as verified, please drop a note here. We'll get this merged. 
Thanks again!

PR #1486, opened 2026-09-06
feat: support b2b_integration_configuration for Enterprise Connect EA

ankita10119:
### 🔧 Changes

Adds support for `b2b_integration_configuration` on clients as part of  the Enterprise Connect EA release

**New field: `b2b_integration_configuration`**
  - Accepted on `POST /api/v2/clients` regardless of `app_type`
  - Contains `integration_type` (one of `custom_auth_server`, `third_party`, `application`) and `sso_profiles` (array of SSO profile IDs)
  - Schema added to the clients handler with full validation

**PATCH guard**
  - The Management API only allows `b2b_integration_configuration` to be updated via PATCH on clients that already have the object; sending it to a client that lacks it returns 400
  - deploy-cli now detects this case, logs a warning, and skips the field, the client is still updated without it
  - To set `b2b_integration_configuration` on an existing client that lacks it, re-create the client

**`app_type: b2b_integration` deprecation**
  - This value was removed from the API in the EA release; the deploy-cli schema has no enum constraint on `app_type` so no code change is needed

**Fix: strip `created_at` and `updated_at` from client PATCH payloads**
  - Discovered during EA testing: the Management API now explicitly rejects these fields with 400 if present in a PATCH body

### 🔬 Testing

 Unit tests added covering:
  - Schema validation accepts valid `b2b_integration_configuration`, rejects invalid `integration_type` values
  - PATCH guard strips the field (with warn) when existing client lacks it
  - PATCH guard strips `null` silently when existing client lacks it
  - PATCH guard passes the field through when existing client already has it

  Manual E2E testing performed against a real tenant:
  - Scenario B (PATCH guard) verified end-to-end - warn log fires, field is skipped, client updates successfully
  - Scenarios A (CREATE) and C (UPDATE round-trip) require `enterprise_connect_entitled` entitlement on tenant; code path confirmed correct, blocked at API level without entitlement

### 📝 Checklist

### DIFF
--- a/src/tools/auth0/handlers/clients.ts
+++ b/src/tools/auth0/handlers/clients.ts
@@ -315,6 +315,25 @@ export const schema = {
         type: 'string',
         description: 'The type of application this client represents',
       },
+      b2b_integration_configuration: {
+        type: ['object', 'null'],
+        description:
+          'B2B integration configuration. Early Access. Required feature flag: enterprise_connect_entitled.',
+        properties: {
+          integration_type: {
+            type: 'string',
+            enum: ['custom_auth_server', 'third_party', 'application'],
+            description: 'The type of integration used to connect to this B2B integration client.',
+          },
+          sso_profiles: {
+            type: 'array',
+            description:
+              'List of SSO profile IDs linked to this B2B integration client. Maximum 1 entry.',
+            items: { type: 'string' },
+          },
+        },
+        additionalProperties: false,
+      },
       resource_server_identifier: {
         type: 'string',
         description:
@@ -663,6 +682,9 @@ export default class ClientHandler extends DefaultAPIHandler {
         'jwks_uri',
         // third_party_security_mode is immutable after creation; PATCH returns 400
         'third_party_security_mode',
+        // API-generated timestamps; PATCH returns 400 if present
+        'created_at',
+        'updated_at',
       ],
       functions: {
         create: (client: Client) => this.createClient(client),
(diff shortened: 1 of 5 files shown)


src/tools/auth0/handlers/clients.ts
```diff
@@ -890,7 +912,31 @@ export default class ClientHandler extends DefaultAPIHandler {
...
+      const payload = { ...client } as Management.UpdateClientRequestContent & {
+        b2b_integration_configuration?: unknown;
+      };
+
+      // b2b_integration_configuration: PATCH only allowed if client already has the object.
+      // Sending it to a client that lacks it returns 400 invalid_body.
+      // null clears the object — but only for clients that already have it; stripping
+      // null for a client without the field is a safe no-op.
+      if ('b2b_integration_configuration' in payload) {
+        const existing = (this.existing || []).find(
```
harshithRai, line 924: This looks right assuming b2b_integration_configuration comes back on the list endpoint, since this.existing is populated from clients.list in getType(). I might be wrong here, but I wasn't sure whether EA fields are actually included in the list response or only on the single-client GET. If they're omitted from list, I think existing?.b2b_integration_configuration would always be falsy and we'd end up silently stripping the field even for clients that legitimately have it, which would quietly break the update round-trip. Could we double check what list returns for an entitled client before we rely on it?

ankita10119: Great catch, totally valid thing to question! Tested this against an entitled tenant,` clients.list` does return `b2b_integration_configuration` with the same shape as the single `GET`, so the guard works correctly. But really appreciate you flagging it, wasn't explicitly verified before!

harshithRai (approved):

PR #1487, opened 2026-09-08
fix: prevent export crash when a handler has no identifiers during keyword preservation (#1446)

### LINKED ISSUE #1446: TypeError: Cannot read properties of undefined (reading 'flat') during export with AUTH0_PRESERVE_KEYWORDS and clientAuthCredentialsPre handler
## Description

When running `a0deploy export` with `AUTH0_PRESERVE_KEYWORDS: true` in the config, the export crashes:

```
error: Problem running command export
error: Problem loading tenant data from Auth0 TypeError: Cannot read properties of undefined (reading 'flat')
```

## Root cause

In `lib/keywordPreservation.js:167`:

```js
const resourceSpecificIdentifiers = auth0Handlers.reduce((acc, handler) => {
    acc[handler.type] = handler.identifiers.flat();  // crashes here
    return acc;
}, {});
```

The `clientAuthCredentialsPre` handler (a deploy-only pre-pass handler registered in `lib/tools/auth0/handlers/index.js:129` via `...{ clientAuthCredentialsPre }`) sets `this.type = 'clientAuthCredentials'` (in `clientAuthCredentialsPre.js:37`) but never sets `this.identifiers` and does not extend any base class that would set it. Therefore `handler.identifiers` is `undefined`, and calling `.flat()` on it throws.

The 8.42.0 release (#1438) fixed a dry-run crash on `clientAuthCredentials`, but the export path through `preserveKeywords` is a separate code path that remains unfixed.

## Suggested fix

Option A — guard against undefined identifiers:
```js
acc[handler.type] = (handler.identifiers || []).flat();
```

Option B — skip handlers without identifiers:
```js
if (handler.identifiers === undefined) return acc;
acc[handler.type] = handler.identifiers.flat();
```

## Reproduction

1. Use any config JSON with `"AUTH0_PRESERVE_KEYWORDS": true`
2. Run `a0deploy export -c config.json --format=yaml --output_folder=./out`
3. Observe crash

The crash occurs during handler iteration in `preserveKeywords`, before any tenant-specific data matters, so any tenant with keyword preservation enabled will trigger it.

## Versions affected

- 8.40.0+ (introduced by #1417 which added `clientAuthCredentialsPre`)
- Confirmed in 8.42.0 (latest) and 9.0.0-beta.1

## Versions not affected

- 8.39.0 and earlier (the `clientAuthCredentialsPre` handler did not exist yet; all handlers had `identifiers` set)

## Minimal reproduction across versions

| Version | Result |
|---------|--------|
| 8.39.0 | passes (no `clientAuthCredentialsPre` handler) |
| 8.40.0+ | crashes (handler exists, `identifiers` undefined) |

harshithRai:
### 🔧 Changes

Exporting with `AUTH0_PRESERVE_KEYWORDS=true` crashed with `TypeError: Cannot read properties of undefined (reading 'flat')` whenever a keyword mapping was set.

`preserveKeywords` builds a map of resource-specific identifiers by reading `handler.identifiers` off every registered handler. Handlers that extend the default `APIHandler` always have this property, but the standalone `clientAuthCredentials` and `clientAuthCredentialsPre` handlers do not, so `handler.identifiers.flat()` threw and aborted the whole export.

This falls back to an empty array when a handler has no `identifiers`, which is the correct value for those handlers since they expose no array-item identifiers to preserve. Export and dry-run now complete normally in both directory and YAML formats.

### 📚 References

- Fixes #1446

### 🔬 Testing

- Added a regression test in `keywordPreservation.test.ts` covering a handler with no `identifiers`.
- Verified against a live tenant: reproduced the original crash on the pre-fix build, then confirmed export succeeds and the keyword marker is preserved, in both directory and YAML formats. Import `--dry-run` completes successfully in both formats.

### 📝 Checklist


src/keywordPreservation.ts
ankita10119, line 249: The runtime fix makes sense! One thing I wanted to check, the `auth0Handlers` parameter type in `preserveKeywords` still declares `identifiers` as required         

Would it make sense to also update this to `identifiers?:` to keep the type in sync with what the code now actually handles? The `@ts-ignore` in the new test seems to exist because of this gap, so I just wanted to flag it in case it was an oversight rather than intentional

src/keywordPreservation.ts
```diff
@@ -254,7 +254,9 @@ export const preserveKeywords = ({
 
   const resourceSpecificIdentifiers: Partial<Record<AssetTypes, string[]>> = auth0Handlers.reduce(
     (acc, handler) => {
-      acc[handler.type] = handler.identifiers.flat();
+      // Standalone handlers (e.g. clientAuthCredentials) don't extend APIHandler, so they
+      // have no `identifiers`; fall back to [] instead of crashing.
+      acc[handler.type] = (handler.identifiers || []).flat();
```
ankita10119: Tiny nit and feel free to ignore, would `??` be preferred over `||` here since the intent is specifically to fall back when identifiers is null or undefined?

 ` acc[handler.type] = (handler.identifiers ?? []).flat();`

test/keywordPreservation.test.ts
```diff
@@ -647,6 +647,27 @@ describe('preserveKeywords', () => {
...
 
+  it('should not crash when a handler has no identifiers', () => {
+    // Regression test for https://github.com/auth0/auth0-deploy-cli/issues/1446
+    const handlersWithMissingIdentifiers = [
+      ...auth0Handlers,
+      //@ts-ignore intentionally omitting `identifiers` to mirror standalone handlers
+      { id: 'id', type: 'clientAuthCredentials' },
+    ];
+
+    expect(() =>
```
ankita10119: The test only verifies the crash is gone. It does not assert that `preserveKeywords` still returns correct keyword-preserved output when a no-identifiers handler is present alongside normal handlers. A stronger regression test would assert the returned value still matches expected output

src/keywordPreservation.ts
harshithRai: done. Made `identifiers` optional so the type matches the runtime behavior, which also let me drop the `@ts-ignore` in the test.

src/keywordPreservation.ts
harshithRai: Agreed, switched to `??` to make the null/undefined-only fallback intent explicit.

test/keywordPreservation.test.ts
harshithRai: Fixed. The test now asserts the full keyword-preserved output with a no-identifiers handler alongside normal ones, not just that it doesn't throw.

ankita10119 (approved):

PR #1492, opened 2026-09-15
 fix: surface real API error on tokenExchangeProfiles 403 responses

ankita10119:
### 🔧 Changes

The tokenExchangeProfiles handler was replacing all 403 responses from the Management API with a hardcoded error message (Token Exchange Profiles feature is not available on this tenant. Please contact Auth0 support to enable this feature.), regardless of the actual cause.                        
         
  
This masked real errors, in particular missing scope errors, making it impossible for customers to diagnose and self-fix the issue.                  
                                                                                                                                                         
The hardcoded string is replaced with the actual API error message, so customers now see the real reason for the failure along with an actionable hint to verify their M2M app scopes and subscription plan. Graceful export behaviour is preserved, the export skips token exchange profiles and continues rather than crashing.

### 🔬 Testing

- Unit tests updated to cover both 403 scenarios: missing scope and feature unavailable on plan
  - Verified on a real tenant:
    - With `read:token_exchange_profiles` scope present → clean export, no regression
    - With scope removed → real API error surfaced in warning, export completes successfully

### 📝 Checklist


test/tools/auth0/handlers/tokenExchangeProfiles.tests.js
```diff
@@ -118,11 +118,28 @@ describe('#tokenExchangeProfiles handler', () => {
...
+        },
+        pool,
+      };
+
+      const handler = new tokenExchangeProfiles.default({ client: pageClient(auth0), config });
+      const data = await handler.getType();
+      expect(data).to.deep.equal([]);
+    });
+
+    it('should return an empty array and warn for 403 with non-scope message', async () => {
```
harshithRai, line 144: One small thing on the tests: they're renamed to "…and warn…", but they only assert `expect(data).to.deep.equal([])`, which also passes against the old code - so the new behavior (surfacing `err.message`) isn't quite covered yet. Would you mind spying on `log.warn` to lock it in? There's a handy precedent in [`scimHandler.tests.js:376-382`](https://github.com/auth0/auth0-deploy-cli/blob/DXCDT-2404/test/tools/auth0/handlers/scimHandler.tests.js#L376-L382):

```js
const warnStub = sinon.spy(log, 'warn');
const data = await handler.getType();
expect(data).to.deep.equal([]);
expect(warnStub.firstCall.args[0]).to.include('Insufficient scope');
warnStub.restore();
```

ankita10119: Addressed and updated

harshithRai (approved):

PR #1493, opened 2026-09-16
fix: handle insufficient_entitlement 403 gracefully for riskAssessment and guardianPolicies handlers 

ankita10119:
## 🔧 Changes

Auth0 now returns `403 insufficient_entitlement` on Adaptive MFA endpoints for tenants without the Enterprise add-on.
Without this fix, export and deploy would crash for those tenants.                                                                                     
   
  **What changed:**                                                                                                                                      
  - Added `isInsufficientEntitlementError` utility to `src/tools/utils.ts` - checks `err.statusCode === 403 && err.body?.errorCode === 'insufficient_entitlement'`
  - `riskAssessmentHandler.getType()` - returns `null` with a warning instead of throwing on `insufficient_entitlement` 403; also handles the case where only `newDevice.get()` 403s
  - `riskAssessmentHandler.processChanges()` - skips with a warning instead of throwing
  - `guardianPoliciesHandler.getType()` - returns `{}` with a warning instead of throwing
  - `guardianPoliciesHandler.processChanges()` - skips with a warning instead of throwing; the `confidence-score` policy triggers this 403 on non-entitled tenants

Non-entitlement 403s (e.g. `insufficient_scope`) are still re-thrown as before.

### 🔬 Testing

 **Unit tests** added for all new code paths in both handlers:
  - `getType()` returns gracefully on `insufficient_entitlement` 403
  - `processChanges()` warns and skips on `insufficient_entitlement` 403
  - Non-entitlement 403s are re-thrown (regression guard)

  **Manual testing** on an entitled tenant 
  - Export: both handlers exported cleanly ✓
  - Deploy with `confidence-score`: applied successfully ✓
  - Idempotency: export → immediate deploy → no changes detected ✓

  **Not tested:** the 403 warn-and-skip path requires a free/non-entitled tenant. This is covered by unit tests. A follow-up ticket should verify end-to-end once a non-entitled tenant is available.

### 📝 Checklist


src/tools/auth0/handlers/guardianPolicies.ts
```diff
@@ -33,9 +34,17 @@ export default class GuardianPoliciesHandler extends DefaultHandler {
...
     if (this.existing) return this.existing;
-    const policies = await this.client.guardian.policies.list();
-    this.existing = { policies };
-    return this.existing;
+    try {
+      const policies = await this.client.guardian.policies.list();
+      this.existing = { policies };
+      return this.existing;
+    } catch (err) {
+      if (isInsufficientEntitlementError(err)) {
```
harshithRai: nit: this catch wraps `policies.list()` (GET), but per the API spec only the PUT set is entitlement-gated. So this export-side branch might probably never get hit.. just a thought

harshithRai (dismissed):

src/tools/auth0/handlers/guardianPolicies.ts
ankita10119: Thanks Harshith
Thats a valid catch, and I have addressed it and updated the PR

harshithRai (approved):

PR #1497, opened 2026-09-17
feat: add handlers for MFA advanced factor config (phone/email OTP + guardian settings)

harshithRai:
### 🔧 Changes

Adds export and import support for MFA Advanced Factor Configuration (Guardian) settings, backed by the endpoints that landed in node-auth0 v7. Three new singleton asset types are introduced, each following the existing Guardian handler pattern (for example `guardianPhoneFactorSelectedProvider`):

- `guardianPhoneFactorSettings`: phone OTP `otp_length` and `otp_expiration_time`, via `guardian.factors.phone.get()`/`.set()`.
- `guardianEmailFactorSettings`: email OTP `otp_length` and `otp_expiration_time`, via `guardian.factors.email.get()`/`.set()`.
- `guardianSettings`: MFA session and "Remember Me" behavior (`display_remember_me_checkbox`, `remember_me_default_value`, `mfa_session_inactivity_timeout`, `mfa_session_overall_timeout`), via `guardian.get()`/`.set()`.

Each asset type ships an API handler plus YAML and directory context handlers, and is registered in the three handler indexes and the `Assets`/`AssetTypes` types. Handlers degrade gracefully when the feature is unavailable on a tenant: a 404 (older Management API) or a forbidden-feature 403 is skipped rather than failing the run. Dry-run is respected, so no writes happen during a preview.

#### Shape

YAML (`tenant.yaml`), before:

```yaml
# no MFA advanced factor config keys were exported
guardianPhoneFactorSelectedProvider:
  provider: auth0
```

YAML (`tenant.yaml`), after:

```yaml
guardianPhoneFactorSettings:
  otp_length: 6
  otp_expiration_time: 300
guardianEmailFactorSettings:
  otp_length: 6
  otp_expiration_time: 300
guardianSettings:
  display_remember_me_checkbox: true
  remember_me_default_value: false
  mfa_session_inactivity_timeout: 604800
  mfa_session_overall_timeout: 2592000
```

Directory format, before:

```text
guardian/
  phoneFactorSelectedProvider.json
  # no settings files
```

Directory format, after, adds three files (`guardian/phoneFactorSettings.json` shown):

```json
{
  "otp_length": 6,
  "otp_expiration_time": 300
}
```

### 🔬 Testing

Verified end-to-end against a live tenant with the feature enabled, in both YAML and directory formats:

- Export produced the three new resources with correct live values.
- Dry-run import reported no changes on an unmodified export (idempotent) in both formats.
- Real import applied a changed phone `otp_length` (6 to 8) in both formats; confirmed by reading the tenant back through the Management API, then reverted to the original value.

Unit tests cover all three resources at each layer (API handler, YAML context, directory context): `getType` handles the 404/403 feature-unavailable cases and rethrows other errors, `processChanges` updates, skips empty assets, and gracefully skips on a 403, and each context handler round-trips parse and dump. Run with `npm test`.

### 📝 Checklist


src/tools/auth0/handlers/guardianEmailFactorSettings.ts
```diff
@@ -0,0 +1,83 @@
...
+  async processChanges(assets: Assets): Promise<void> {
+    // No API to delete or create guardianEmailFactorSettings, we can only update.
+    const { guardianEmailFactorSettings } = assets;
+
+    // Do nothing if not set
+    if (!guardianEmailFactorSettings) {
+      return;
+    }
+
+    if (isDryRun(this.config)) {
```
ankita10119, line 60: Looks like this guard only returns early when there are no changes, if `update.length > 0` (e.g. user changed `otp_length`), doesn't it fall through and call set() even in dry-run? Same pattern on **guardianEmailFactorSettings.ts:58–65** and **guardianSettings.ts:70–77**. Would something like this work?        
                                                                                                                                                         
```
  if (isDryRun(this.config)) {                                                                                                                           
    await this.calcChanges(assets);                                                                                                                      
    return;                                                                                                                                              
  }
```

test/tools/auth0/handlers/guardianEmailFactorSettings.tests.js
```diff
@@ -0,0 +1,210 @@
...
+      const stageFn = Object.getPrototypeOf(handler).processChanges;
+
+      // Should resolve (not throw) even though the API rejects with a 403.
+      await stageFn.apply(handler, [
+        { guardianEmailFactorSettings: { otp_length: 6, otp_expiration_time: 300 } },
+      ]);
+      expect(handler.updated).to.equal(0);
+    });
+
+    it('should not write when in dry-run mode and there are no changes', async () => {
```
ankita10119, line 158: This dry-run test covers the no-change case, would it make sense to also add one where the local value differs from remote (e.g. `otp_length: 8` vs the
`mocked get()` returning 6), just to make sure `set()` isn't called there either?

harshithRai: Added coverage for the changed-value path in all three suites. Since `processChanges` in dry-run only runs in apply-after-preview, the test asserts `set()` is called there rather than skipped.

src/tools/auth0/handlers/guardianEmailFactorSettings.ts
harshithRai: `processChanges` isn't reached during a pure dry-run, `deploy.ts` runs the preview via `dryRun()`/`dryRunChanges()` and returns before `processChanges` unless the user opts to apply. It only runs here in the `apply-after-preview` path (`AUTH0_DRY_RUN_APPLY` or interactive "Apply changes"), where `AUTH0_DRY_RUN` is still set and writing is intended. The guard just skips the no-op `set()` when nothing changed. Same as `emailProvider`/`attackProtection`/`guardianPolicies`. `calcChanges` + `return` would break `apply-after-preview`, so leaving as is.

ankita10119 (approved):
LGTM

PR #1500, opened 2026-09-18
feat: add support for auth_email_by_code email template

EmilioCristalli:
Auth0's SDK already exposes this template name; wire it into the supported email template types/names so import/export and validation recognize it.

### 🔧 Changes

Add support for the `auth_email_by_code` email template.

### 📚 References

Management API already supports it https://auth0.com/docs/api/management/v2/email-templates/get-email-templates-by-template-name

### 🔬 Testing

Tested by exporting from an existing tenant with `node lib/index.js export`

### 📝 Checklist


harshithRai:
Thanks for putting this together, really nice addition to have `auth_email_by_code` supported.

I gave it a test run and the feature side looks good: the template returns a 404 when it isn't configured (not a 403), so export handles it without issues, and a dry run import validates and picks it up fine.

The only thing holding CI back is some stale test fixtures. Now that `auth_email_by_code` is in `EMAIL_TEMPLATES_TYPES`, the YAML dump test emits one extra template, but the three expected `emailTemplates` arrays in `test/context/yaml/context.test.js` still list the old set. That's what's tripping the three failing cases (`should dump tenant.yaml with defaults`, `without defaults`, and `with INCLUDED and EXCLUDED props including defaults`).

If you add this to each of those three arrays, right after the `async_approval` block so it stays alphabetical, they should go green:

```js
{
  body: './emailTemplates/auth_email_by_code.html',
  enabled: true,
  template: 'auth_email_by_code',
},
```

I ran the full suite with that change locally and everything passed. Appreciate the contribution!

harshithRai (changes requested):
good PR, pending the test fixes.
Please drop a note here when you're done with this so, we can proceed with the approval and merge!

EmilioCristalli:
thanks for the review @harshithRai !
i think i fixed the tests, but there's still an e2e test failing. Should i also look into that or do you guys manage e2e tests in a different way?

harshithRai:
> thanks for the review @harshithRai ! i think i fixed the tests, but there's still an e2e test failing. Should i also look into that or do you guys manage e2e tests in a different way?

Thanks for the quick turnaround on the fixtures. CI is green now (the E2E job failing is expected for PRs from a fork, nothing to worry about). I tested export and import against a live tenant and it behaves correctly. Looks good to me, approving.

Kudos to your contribution 🫡

harshithRai (approved):

